PR #11, opened 2019-12-20
[MAJOR RFC] drop LGPL 2.0 license option

brodybits:
(in package.json)

for the sake of simplicity

TODO:

- [ ] fix the LICENSE file (it should maybe contain the text of the MIT license)
- [ ] this should probably be in a new 0.x release since it should be considered a breaking change

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,5 +56,5 @@
     "email": "jindw@xidea.org",
     "url": "http://github.com/xmldom/xmldom/issues"
   },
-  "license": "(LGPL-2.0 or MIT)"
+  "license": "MIT"
 }


timbru31:
For reference: https://tldrlegal.com/license/gnu-lesser-general-public-license-v2.1-(lgpl-2.1)

- IMO before we can re-license the code we would need the OK from every past contributor. But I'm no legal expert.

brodybits:
I thought it should be OK to just drop a license option, since it is or should be an option. But maybe easiest to just leave it as-is and close this PR. Any comments from @kethinov?

brodybits:
from @kethinov: https://github.com/xmldom/xmldom/pull/13#issuecomment-568800153

closing now

PR #25, opened 2020-03-01
Fix #24 Implement Document.getElementsByClassName

### LINKED ISSUE #24: Implement Document.getElementsByClassName
Can we add [Document.getElementsByClassName](https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementsByClassName)?

Currently Im doing this workaround to polyfill getElementsByClassName
```javascript
const patchDoc = doc => {
  doc.getElementsByClassName = search => {
    const results = [];
    const elements = doc.getElementsByTagName('*');
    const pattern = new RegExp(`(^|\\s)${search}(\\s|$)`);
    for (let i = 0; i < elements.length; i++) {
      if (pattern.test(elements[i].getAttribute('class'))) {
        results.push(elements[i]);
      }
    }
    return results;
  };
  return doc;
}

const doc = patchDoc(new DOMParser().parseFromString(html));
```

codler:


dom.js
```diff
@@ -608,6 +608,21 @@ Document.prototype = {
 		return rtv;
 	},
 	
+	getElementsByClassName: function(className) {
+		const pattern = new RegExp(`(^|\\s)${className}(\\s|$)`);
```
kethinov: Try merging in the latest master to your PR and re-running the tests to see if Node 10.x no longer complains about this line.

PR #33, opened 2020-03-04
cleanup: remove .project

brodybits:
should be a new 0.x release such as 0.4.0


kethinov:
I didn't remove that file in my cleanup PR because I wasn't sure what it was for. Any idea?

brodybits:
Seems to have been added for some kind of a test sample in 2943798f56d56922b04f6bb7701f981369198095

I think it should be OK to remove it in a new 0.x release such as 0.4.0. I just created a new "0.4.0" milestone and added this PR to that milestone.

Issue #35, opened 2020-03-04
test coverage seems to be very limited


brodybits:
`npm test` continues to succeed if I make some non-sense edits in my work area:

```diff
% git diff      
diff --git a/lib/dom-parser.js b/lib/dom-parser.js
index 92937fd..1726488 100644
--- a/lib/dom-parser.js
+++ b/lib/dom-parser.js
@@ -246,6 +246,6 @@ function appendElement (hander,node) {
 var htmlEntity = require('./entities');
 var XMLReader = require('./sax').XMLReader;
 var DOMImplementation = exports.DOMImplementation = require('./dom').DOMImplementation;
-exports.XMLSerializer = require('./dom').XMLSerializer ;
-exports.DOMParser = DOMParser;
+//exports.XMLSerializer = require('./dom').XMLSerializer ;
+//exports.DOMParser = DOMParser;
 //}
diff --git a/lib/dom.js b/lib/dom.js
index 01794c8..5ffcf27 100644
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -1250,7 +1250,7 @@ try{
 }
 
 //if(typeof require == 'function'){
-       exports.Node = Node;
-       exports.DOMImplementation = DOMImplementation;
-       exports.XMLSerializer = XMLSerializer;
+       //exports.Node = Node;
+       //exports.DOMImplementation = DOMImplementation;
+       //exports.XMLSerializer = XMLSerializer;
 //}
```

I think we need to massively improve the test suite before we can do much more cleanup work.

brodybits:
I have been using Stryker Mutator on some other projects, highly recommended we consider using it for xmldom.

brodybits:
I would personally be very open to switching from proof to a well-known test library or framework. I generally like to use Jest in new projects since it just works for both normal JavaScript projects and React Native. I would also not mind adding the new test library before dropping use of the existing `proof` testlib. This could help unblock testing for PRs like #41.

karfau:
I'm very interested in supporting this. any specific tasks/what would be a good first test to have (in a new test framework)?
I have some experience with `mocha`, `tap` and `jest` but none with `proof`.

brodybits:
I would start with adding test coverage for the cases in the description.

I think it would be ideal if someone could get Stryker Mutator working with this project.

[...]

PR #36, opened 2020-03-04
Cleaner module exports with index.js

brodybits:
in followup to the maintenance done in PR #28 for 0.3.0

I think we should add the missing test coverage discussed in #35 before we consider merging this proposal.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,9 @@
+const { DOMParser } = require('./lib/dom-parser')
+
+const { DOMImplementation, XMLSerializer } = require('./lib/dom')
+
+module.exports = {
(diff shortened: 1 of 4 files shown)


karfau:
@brodybits since it is marked as a draft I think it doesn't need to be in the upcoming 0.6.0 release, right?

karfau:
@brodybits I think there are some issues with this PR, like usage of `const` inside `lib`.
There are also some conflicts that would need to be resolved and I believe that this PR has been "made obsolete" by #233.

So I hope it is OK to close this PR.

If you disagree please reopen and let me know why.

PR #41, opened 2020-03-16
Additional public Level 2 DOM interfaces

### LINKED ISSUE #40: Expose DOM Level 2 Elements
Glad you all have forked and are maintaining `xmldom`, thank you!

There were [several](https://github.com/jindw/xmldom/pull/144) [issues](https://github.com/jindw/xmldom/issues/235) in the old unmaintained repo dealing with exposing DOM Level 2 object interfaces, such as `Node`, `Element`, `NodeLike`. Right now, those classes are private.

There are multiple library that deal with the browser `DOMParser` that make use of these for things like `instanceof` checks. As one example, I'm using `bs-xml`, which wraps `DOMParser` for Bucklescript/ReasonML, and which  [has to check](https://github.com/zpnester/bs-xml/blob/a5c553dd56563f9f3eb31fdbc0f107394df73560/src/Xml__.re#L7) if an object is a `Node` or `Element`.

Would you be open to a very simple PR that exposes those classes? Specifically, the ones listed as fundamental interfaces in [the Level 2 DOM Core spec](https://www.w3.org/TR/2000/REC-DOM-Level-2-Core-20001113/core.html)?

```
DOMException
ExceptionCode
DOMImplementation
DocumentFragment
Document
Node
NodeList
NamedNodeMap
CharacterData
Attr
Element
Text
```

esbullington:
Make all the Level 2 Core DOM interfaces public.

Closes #40

### DIFF
--- a/lib/dom-parser.js
+++ b/lib/dom-parser.js
@@ -1,3 +1,8 @@
+var htmlEntity = require('./entities');
+var XMLReader = require('./sax').XMLReader;
+var DOM = require('./dom');
+var DOMImplementation = DOM.DOMImplementation;
+
 function DOMParser(options){
 	this.options = options ||{locator:{}};
 }
@@ -242,10 +247,20 @@ function appendElement (hander,node) {
     }
 }//appendChild and setAttributeNS are preformance key
 
-//if(typeof require == 'function'){
-var htmlEntity = require('./entities');
-var XMLReader = require('./sax').XMLReader;
-var DOMImplementation = exports.DOMImplementation = require('./dom').DOMImplementation;
-exports.XMLSerializer = require('./dom').XMLSerializer ;
+// Primary Web API interfaces
 exports.DOMParser = DOMParser;
-//}
+exports.XMLSerializer = DOM.XMLSerializer;
+// Level 2 Core DOM interfaces:
+// https://www.w3.org/TR/2000/REC-DOM-Level-2-Core-20001113/core.html
+exports.DOMImplementation = DOMImplementation;
+exports.Node = DOM.Node;
+exports.NodeList = DOM.NodeList;
+exports.Element = DOM.Element;
+exports.NamedNodeMap = DOM.NamedNodeMap;
+exports.DOMException = DOM.DOMException;
+exports.ExceptionCode = DOM.ExceptionCode;
+exports.DocumentFragment = DOM.DocumentFragment;
+exports.Document = DOM.Document;
+exports.CharacterData = DOM.CharacterData;
+exports.Attr = DOM.Attr;
+exports.Text = DOM.Text;
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -1249,8 +1249,9 @@ try{
 }catch(e){//ie8
 }
 
-//if(typeof require == 'function'){
-	exports.Node = Node;
-	exports.DOMImplementation = DOMImplementation;
-	exports.XMLSerializer = XMLSerializer;
-//}
+exports.NamedNodeMap = NamedNodeMap;
+exports.NodeList = NodeList;
+exports.Element = Element;
+exports.Node = Node;
+exports.DOMImplementation = DOMImplementation;
+exports.XMLSerializer = XMLSerializer;


lib/dom-parser.js
```diff
@@ -1,3 +1,8 @@
+var htmlEntity = require('./entities');
+var XMLReader = require('./sax').XMLReader;
+var DOM = require('./dom');
+var DOMImplementation = DOM.DOMImplementation;
```
brodybits: I would *personally* rather see newly added code use `const` declarations whenever possible, which should be no problem on the now minimum Node.js version of 10.

lib/dom-parser.js
```diff
@@ -242,10 +247,20 @@ function appendElement (hander,node) {
...
-//if(typeof require == 'function'){
-var htmlEntity = require('./entities');
-var XMLReader = require('./sax').XMLReader;
-var DOMImplementation = exports.DOMImplementation = require('./dom').DOMImplementation;
-exports.XMLSerializer = require('./dom').XMLSerializer ;
+// Primary Web API interfaces
 exports.DOMParser = DOMParser;
-//}
+exports.XMLSerializer = DOM.XMLSerializer;
+// Level 2 Core DOM interfaces:
```
brodybits: [Cleanup] I would add a space here, for the sake of easier readability

```suggestion
exports.XMLSerializer = DOM.XMLSerializer;

// Level 2 Core DOM interfaces:
```

lib/dom-parser.js
esbullington: I agree 100%, but I noticed the entire rest of the codebase was using ES5. That's why I kept it consistent.

lib/dom-parser.js
esbullington: Do you want me to clean up the formatting? The existing code base had mixed tabs and spaces, and I initially cleaned all that up, but was concerned it would make the PR harder to review.

lib/dom-parser.js
brodybits: @esbullington I think you are right to keep it consistent at this point.

brodybits:
Thanks @esbullington for your work on this, sorry it is not as easy as we would like. Leaving the styling up to you.

I do think the lack of test coverage that I raised in #35 is a major unfortunate blocker.

I am adding this to the 0.5.0 milestone for now. Maybe we can get this done sooner.

esbullington:
Yeah, no problem about the tests, it's wise. I discovered what you did in #35 while trying to figure out the test system.

I can easily add tests for my PR (easy enough to test equality of parsed node, etc via `instanceof` which is exactly how I'm using them). But not sure about the testing, I've never used this `proof` library before and just trying to figure it out.

brodybits:
> But not sure about the testing, I've never used this `proof` library before and just trying to figure it out.

See <https://github.com/xmldom/xmldom/issues/35#issuecomment-601831429>: I would be very open to using a new test library such as Jest for new work and to start migrating away from `proof`.

lib/dom-parser.js
```diff
@@ -242,10 +247,20 @@ function appendElement (hander,node) {
...
+// https://www.w3.org/TR/2000/REC-DOM-Level-2-Core-20001113/core.html
+exports.DOMImplementation = DOMImplementation;
+exports.Node = DOM.Node;
+exports.NodeList = DOM.NodeList;
+exports.Element = DOM.Element;
+exports.NamedNodeMap = DOM.NamedNodeMap;
+exports.DOMException = DOM.DOMException;
```
karfau, line 260: This was already exported (and documented as part of #174

karfau:
> But not sure about the testing, I've never used this proof library before and just trying to figure it out.

@esbullington we moved to jest, so it should be easier to get started now

karfau:
This one also doesn't look like it's ready for the current release.

karfau:
@esbullington Thank you for investing the time, your contribution is welcome, it just had really bad luck regarding the timing. The last months have been quite a rough ride for this library and it's maintainers. We are trying to improve the transparency on which issues we try to resolve when and when contributions make sense.

I think the approach you are taking is a valid one, but we will need to resolve the conflicts and add tests.
But I also asked a question in #40 wich we should resolve.

Even when all of this is resolved I'm hesitant to merge this change right now, because it would make sense to first land #233, so that those exports do not have to be in `lib/dom-parser.js` (where they are not used), but instead can be in `lib/index.js` (not yet existing).

Because of all of this I removed the milestone from this PR and also decided to close this PR.
This was not an easy decision, but I think we will make better use of all of our time this way.

PR #43, opened 2020-03-30
Replace regex with an equivalent using String.indexOf()

matthewfleming:
Hi there,

I work for Seven West Media. We recently had an incident on our content aggregator service - a node server that ingests news stories from our partners. It uses xmldom to parse the stories. There was a particular package that was causing node to crash, and block other packages from processing. We narrowed the problem down to line 69 of sax.js - the `Regex.exec()`.

Our solution was to swap out the regex for a String.indexOf().

For legal reasons I can't actually send you the xml in question (470K, 2 lines, valid xml), but here is a quick profile of the regex vs indexOf versions:
```js
# test.js
var DOMParser = require('xmldom').DOMParser;
var fs = require('fs');
var data = fs.readFileSync('B881498861Z.1_20200325063359_000.xml', 'utf8');
var doc = new DOMParser().parseFromString(data);
console.log('done');
```
```
➜  xmldom git:(master) ✗ time node test.js
done
node test.js   39.12s  user 0.11s system 99% cpu 39.253 total

➜  xmldom git:(bugfix/replace-crashy-regex2) ✗ time node test.js
done
node test.js   0.17s  user 0.03s system 123% cpu 0.164 total
```

### DIFF
--- a/lib/sax.js
+++ b/lib/sax.js
@@ -65,12 +65,15 @@ function parse(source,defaultNSMapCopy,entityMap,domBuilder,errorHandler){
 			start = end
 		}
 	}
-	function position(p,m){
-		while(p>=lineEnd && (m = linePattern.exec(source))){
-			lineStart = m.index;
-			lineEnd = lineStart + m[0].length;
+	function findLineEnd(){
+		var newLine = source.indexOf('\n', lineEnd);
+		return newLine === -1 ? source.length : newLine + 1;
+	}
+	function position(p){
+		while(p>=lineEnd){
+			lineStart = lineEnd;
+			lineEnd = findLineEnd();
 			locator.lineNumber++;
-			//console.log('line++:',locator,startPos,endPos)
 		}
 		locator.columnNumber = p-lineStart+1;
 	}


Epsiloncool:
Here is a good article which described the reason of the issue.

[](https://javascript.info/regexp-catastrophic-backtracking)

I spent 2 days to find up the reason and finish with my own solution.

brodybits:
A test case would really help. The change does seem to improve efficiency, though.

matthewfleming:
> Thank you for providing the change that works on your example.
> But I don't think it's a good idea to drop the detection or `\r\n?|` as of now since we know we are having #49 and I just checked that [we have no tests for certain parts of the related regular expressions](https://github.com/karfau/xmldom/commit/5356ffc8166e5dff154d4b04180846b7186b7f63)

I'm not sure these things are necessarily related - the spec quoted in the issue says you should preprocess the xml to remove the \r before parsing.

> I agree a fix for this has a high prio but we should of course avoid to break other things for it.

I agree it needs a test. I needed to move on to other tasks at the time but I'll look to add one.

karfau:
For reference: Issue #97 is describing the same case and offers a test case as an attachment.

matthewfleming:
I am no longer able to reproduce the error with my original xml file or the linked one. I can only assume a bug has been patched in node or a system library that was causing the issue.

PR #52, opened 2020-06-20
build: Add `.nvmrc`

karfau:
To allow easy version switching to the version configured for CI.

https://github.com/nvm-sh/nvm#nvmrc


brodybits:
Definitely looks like an interesting idea. I generally like to use <https://github.com/jasongin/nvs> for Node.js version switching which seems to work the same way on "*nix", Windows cmd, and PowerShell.

karfau:
Cool, I just checked: `nvs` also supports `.nvmrc` file. And it will pick it up when `nvs auto on` has been run.

brodybits:
I guess this makes sense, never used this feature myself though. I think GitHub CI would generally test with all supported Node.js versions. With nvs, I generally just use 1-2 recent Node.js versions with a few basic tools like Yarn installed and ready to go. I think it would be ideal to have one opinion on this one.

timbru31:
We use a `.node-version` file at work - IIRC this file is recognized by more "automatic version switchers" than a `.nvmrc`.

brodybits:
> We use a `.node-version` file

Nice idea. Both files seem to be honored by nvs, `.node-version` does not seem to be honored by nvm ref:

- https://github.com/nvm-sh/nvm/issues/794
- https://stackoverflow.com/a/29545541/1283667
- https://github.com/jasongin/nvs#automatic-switching-per-directory

It looks like they did start discussion of standardizing `.node-version` format in <https://github.com/nodejs/version-management/issues/13>, seem to be a bit long running.

While I am pretty biased towards nvs over nvm, I would favor maximum interoperability whenever possible.

timbru31:
Ah indeed, I use https://github.com/wbyoung/avn-nvm, too, to support the `.node-version` file.

brodybits:
I noticed the file was renamed, maybe the title should be updated. But I am not yet convinced this is what we want.

I tried to imply that this renaming to `.node-version` would not work with the extremely popular Node.js version tool (nvm), which is not very good, and it looks like @timbru31 agrees. Any more comments from @timbru31?

karfau:
Oh, sorry. I misread the comments, you are right @brodybits .

Just thinking out loud here: It sounds aweful, but I could of course create a link from `.node-version` to `.nvmrc`, so they always contain the same information...

Just read through (most of) those year long threads and I really don't want to continue the discussion here.
If there is no common agreement for this file format yet (also not just between us) I would suggest we make it something that everybody needs to configure personally.
E.g. I will add that file to my global `.gitignore` and still keep it in my local copy.

Feel free to close the PR if you think this topic is just not important(/doable) enough right now.

brodybits:
> create a link from `.node-version` to `.nvmrc`

We could do it the other way around, like this:

- keep it in `.node-version`
- do `ln -s .node-version .nvmrc`
- do `git add .nvmrc` and commit

My rationale is that nvm is what needs the `.nvmrc` assumes shell scripts, and hopefully we can just remove it someday.

> Just read through (most of) those year long threads and I really don't want to continue the discussion here.

I sure don't blame you:)

> add that file to my global `.gitignore` and still keep it in my local copy

I would actually favor that kind of solution. I would personally not likely use this feature, as I said before. Thanks for looking into this, though.

karfau:
Let's rather invest our time into different topics. :handshake:

karfau:
I just found out about the following documentation repo that was created 7 days after we closed this PR (and is still maintained):
https://github.com/shadowspawn/node-version-usage#suggested-compatible-format

So it looks like we would be able to have `.node-version` with the "suggested compatible format" to "pin" the the default version to use during development.

PR #51, opened 2020-06-20
Restore `vows` test "suite", drop useless proof tests

karfau:
fixes #35 

I found that some part of the `test` folder are using [`vows`](https://github.com/vowsjs/vows) for testing.
- Main changes in the test files is for fixing imports and removing "traversal code" that `vows` is able to handle when invoked from the cli. (`test/**/.js` files that don't have the `.vows.js` extension might still contain assertions. We need to digest them and decide whether and how to restore them.)
- To ease filtering I '''renamed''' those files in an extra commit. we should keep the commits when landing so the git history of those files is not lost. (**For reviewing it's also much easier to go commit by commit.**)
 
All the `vows` tests can now be executed via the usual `npm test`.
Tons of test cases where only calls to `console.assert` that logged an "Assertion failed: ...", I converted them into `assert.[equal|isTrue|greater|...]` where possible. When these failed and it made sense, I changed the expectation to the current value.
In case we need to skip a test, the easiest is still to convert it to `console.assert`.

I also drastically reduced the amount of `console.log` in tests, now the only thing that's printing in addition to `vows` is the internal logging of xmldom.

Stryker can be run using `npm start` for now, I thought it's the most convenient solution for now. See below for the report from my first run.

- I think we can add `jest` in a different PR.
- I found that `&nbsp;` is converted to a normal space instead of `\u00a0`, see [comment](#issuecomment-647081281) and commit afterwards that fixes the regression.

Issue #57, opened 2020-06-22
Regression regarding handling of `&nbsp;` with "(x)html" mimeType


karfau:
As [discovered in #51](https://github.com/xmldom/xmldom/pull/51#issuecomment-647081281), when parsing `<body>&nbsp;word</body>` with one of the possible "(x)html" mimeTypes, already in the parsed object structure `nbsp;` has been replaced with a normal (breaking) space. When it's serielized it then returns `<body xmlns="http://www.w3.org/1999/xhtml"> word</body>`.

The regression was introduced in 6b47d09be5aca3e0b4687812dfccf452482bc660 ([that handels some cases correctly](https://github.com/xmldom/xmldom/commit/6b47d09be5aca3e0b4687812dfccf452482bc660#diff-b7abc4fd0b8aebd87eb1b0fce7e40a45R270-R272)) and it's impact was widened 56eb39f82dd2052a683b1870fb1f4a105e184f66
and first published as part of v0.2.0 ([github](https://github.com/xmldom/xmldom/releases/tag/0.2.0), [npm](https://www.npmjs.com/package/xmldom/v/0.2.0))

A fix is available in 908b9da2 but we want to first fix #35 and #56

brodybits:
> The regression was introduced in [6b47d09](https://github.com/xmldom/xmldom/commit/6b47d09be5aca3e0b4687812dfccf452482bc660) ([that handels some cases correctly](https://github.com/xmldom/xmldom/commit/6b47d09be5aca3e0b4687812dfccf452482bc660#diff-b7abc4fd0b8aebd87eb1b0fce7e40a45R270-R272))

Looking at the changes, I have a feeling the `&nbsp;` functionality would not have worked at all before the changes were applied.

karfau:
> > The regression was introduced in [6b47d09](https://github.com/xmldom/xmldom/commit/6b47d09be5aca3e0b4687812dfccf452482bc660) ([that handels some cases correctly](https://github.com/xmldom/xmldom/commit/6b47d09be5aca3e0b4687812dfccf452482bc660#diff-b7abc4fd0b8aebd87eb1b0fce7e40a45R270-R272))
> 
> Looking at the changes, I have a feeling the `&nbsp;` functionality would not have worked at all before the changes were applied.

[...]

PR #59, opened 2020-06-26
Recover `vows` tests, drop `proof` tests

### LINKED ISSUE #35: test coverage seems to be very limited

karfau:
fixes #35 

I found that some part of the `test` folder are using [`vows`](https://github.com/vowsjs/vows) for testing.
- Main changes in the test files is for fixing imports and removing "traversal code" that `vows` is able to handle when invoked from the cli. (`test/**/.js` files that don't have the `.vows.js` extension might still contain assertions. We need to digest them and decide whether and how to restore them.)
- To ease filtering I '''renamed''' those files in an extra commit.
- As discussed in #51 some files still contain CRLF line separators

All the `vows` tests can now be executed via the usual `npm test`.
Tons of test cases where only calls to `console.assert` that logged an "Assertion failed: ...", I converted them into an equivalent assertion that fails a test.
Due to the discussion in the first PR I introduced a helper module [`test/assert.js` that is heavily covered by tests](https://github.com/xmldom/xmldom/pull/59/commits/4f24b5fadd6311a5a0d081b82c865dc220f9bed7) and
- allowed me to more easily restore existing assertions
- allows us to switch between loose `==` and strict `===` checking by `XMLDOM_ASSERT_STRICT=1 npm run test`

When these failed and it made sense, I changed the expectation to the current value.
In case we need to skip a test, we can use `assert.skip` (that logs the failed assertion and fails if the assertion no longer fails).

I also drastically reduced the amount of `console.log` in tests, now the only thing that's printing in addition to `vows` is the internal logging of xmldom.

**For reviewing it's much easier to go commit by commit!** (Watching all changes in one go will look like history of the test files gets lost, but this is not true **as long as commits are preserved when the PR lands**.)


brodybits:
I may need a few days to review. I have some quick questions:

- Is the code in `test/assert*.js` adapted from somewhere?
- I wonder if any the code in `test/assert*.js` could be moved into a new package that could help some others?
- 37d5fb9dc2dad9670a59ec76d9a64d483d3ec199 - "Assert the behavior since 0.1.27" - looks like a possible regression that happened in a patch release. Should we raise this as an issue?

karfau:
> * Is the code in `test/assert*.js` adapted from somewhere?

No, [the question whether moving a test suite from loose asserts to strict asserts could allow future breaking changes](https://github.com/xmldom/xmldom/pull/51#pullrequestreview-434518082) got me thinking and I made me write this, so we can observe and treat this question in a central place.

> * I wonder if any the code in `test/assert*.js` could be moved into a new package that could help some others?

I have written some assertion helpers before, and this might be the most complex piece so far.
Most of the time they have one thing in common: project specific assumptions. 
This one has to and it allows for a great deal of reduced complexity. E.g.
- using `vows` to write tests
- using node's `assert` to write tests and implement it
- using projoect specific folder structure to locate the actual assertion in test code
- the name of the env var to set
- the assumption or rather decision that a skipped assertion that no longer fails should fail the test
- throwing when any caught Error is not an `/^AssertionError/`
- which "native" node assertions to provide on it
- ...

So currently I'll not invest time into that, but I'm planning to create an article about that question on dev.to and we will see how it evolves from there.

> * [37d5fb9](https://github.com/xmldom/xmldom/commit/37d5fb9dc2dad9670a59ec76d9a64d483d3ec199) - "Assert the behavior since 0.1.27" - looks like a possible regression that happened in a patch release. Should we raise this as an issue?

I went ["back in time" as described here](https://github.com/xmldom/xmldom/pull/51#issuecomment-647081281) again (with the tests from this branch) and discovered:

- the test in `xss-test.vows.js` works in 0.1.27
- the [`assert.throws` in unclosedcomment.vows.js](https://github.com/xmldom/xmldom/pull/59/commits/37d5fb9dc2dad9670a59ec76d9a64d483d3ec199#diff-21edcb06b126f14618f77a5b2bfb0e20L9-L12) succeds because `dom.toString` fails (not when parsing):
```
      ✗ unclosedcomment 
      TypeError: Cannot read property 'prefix' of null 
      at Document.nodeSerializeToString [as toString] (xmldom/dom.js:920:23) 
      at Object.unclosedcomment (xmldom/test/parse/unclosedcomment.vows.js:10:10) 
      at runTest (xmldom/node_modules/vows/lib/vows.js:136:26) 
      at EventEmitter.<anonymous> (xmldom/node_modules/vows/lib/vows.js:81:9) 
      at EventEmitter.emit (events.js:203:15) 
      at EventEmitter.options.Emitter.emit (xmldom/node_modules/vows/lib/vows.js:241:24) 
      at xmldom/node_modules/vows/lib/vows/suite.js:170:45 
      at process._tickCallback (internal/process/next_tick.js:61:11) 
      at Function.Module.runMain (internal/modules/cjs/loader.js:834:11) 
      at startup (internal/bootstrap/node.js:283:19) 
```
which was fixed [here](https://github.com/xmldom/xmldom/commit/56eb39f82dd2052a683b1870fb1f4a105e184f66#diff-7d1c5d97786fdf9af5446a241d0b6d56L919-R919) which I would consider a bugfix (released in 0.2.0) and not a breaking change.


Just to be sure we are on the same page, here is my basic assumption: 
> Since v 0.1.27 was published 3.5 years (-2 days) ago by the original maintainer I would generally consider this the "version people rely on by now". 
> 
> So fixing things that could be assumed before that would "only" help people that stuck to an even earlier version, I'm not sure we want to already discuss those problems but we can easily run those tests against earlier versions to detect those and discuss which behavior makes more sense.
> 
> I would recommend to only do that after we have a more robust/complete test suite.

Please let me know in case you disagree.

brodybits:
Thanks. Your disposition concerning the asserts makes more sense to me now. I will try to take a closer look this weekend, don't expect to find any more blockers at this point. But I think it would be great if we could raise issues for the following:

- improving the asserts
- the changed behavior from 0.1.27
- maybe an issue to track your checklist from #57

test/assert.js
```diff
@@ -0,0 +1,114 @@
...
+}
+
+/**
+ * Picks the location of the assertion from the stacktrace if available.
+ *
+ * @param {Error} error
+ */
+function locateAssertion (error) {
+  return error && error.stack
+    && error.stack.split('\n').find(
```
brodybits, line 14:
```suggestion
  return error && error.stack &&
    error.stack.split('\n').find(
```

brodybits: We could also put the whole return expression in parens.

In general, I would like to avoid potential confusion due to the way JavaScript handles return expressions (which I think I do not understand 100%).

I think the Prettier tool would follow my suggestion, with the break after binary operator. (Prettier does put the break before ternary operators, which took me a while to get used to. They seem to be considering changing the split for binary operators in a future release, but not so easy to get sufficient level of consensus.)

brodybits:
I did also try running Stryker on `test/assert.js`. It did encounter some timeouts but did not find any surviving mutants, which gives me some extra confidence. I would like to add Stryker after merging this one, see #60.

test/assert.js
karfau: I think this change is just about formatting and from that perspective it's just a tiny part of it. 
Let's introduce prettier in some later PR and fix all those at once.
If you think it helps just apply the change, I gave you edit rights to the branch when I created the PR.

karfau:
Regarding the checklist in #51 (

> * improving the asserts

If you are referring to using strict asserts everywhere I can do it, if you have more in mind, maybe you should create that issue.

> * the changed behavior from 0.1.27

You mean creating issues just to close them? Or one to list all of them? I'm not sure what to create an issue for. So maybe it's easier for you to do it?

> * maybe an issue to track your checklist from #57

Are you referring to [this checklist](https://github.com/xmldom/xmldom/pull/51#issuecomment-647298093)?
I would prefer to maintain it there, as far as I can see all items that are relevant to this repo are already either issues or PRs. So maybe I misunderstood you again...

Thx for running stryker against the `test/assert.js` nice idea (I thought about it, but didn't try)
Regarding adding stryker in this PR:
I think the mixed line separator issue is the most burning one, since it's making it hard to contribute.
I would prefer for myself to get the checklist from the first PR done, next thing can be Stryker, I don't care about the credits. But let's not add more things to this PR, I would say it's big enough.

brodybits:
I just raised #61 to improve the test assertions.

> > * the changed behavior from 0.1.27
> 
> You mean creating issues just to close them? Or one to list all of them?

Let's start with one to list them all for now, but let's keep the regression in #57 open. My purpose is to avoid losing track.

> Are you referring to [this checklist](https://github.com/xmldom/xmldom/pull/51#issuecomment-647298093)?
> I would prefer to maintain it there

In that case, my idea would be to keep it open as a draft PR, maybe put something in brackets at the beginning of the title. Again my purpose is to avoid losing track.

I will do a final review and merge this weekend. Please do me a favor and ping me, multiple times if needed, just in case I drop it for any reason. I am definitely with you to resolve the line endings and finish that checklist.

For the record, I think this PR needs to be merged in a merge commit. But I think this should be an exception, general preference should be squash commit.

And thanks again for your contributions!

karfau:
Since you asked for it, @brodybits here is your first ping.
How is it going? :wink:

brodybits:
Pong, thanks!

I will probably make the formatting edit, one last review & test, and merge this one.

PR #64, opened 2020-06-28
add test-node.js.yml to GitHub workflows

brodybits:
from template on GitHub UI

with the following changes:

- update filename
- update name field
- replace `npm ci` with `npm install`
- remove `npm run build` step

This *should* get CI testing working again both on master branch and in PRs

This is in followup to PR #59

/cc @karfau

### DIFF
--- a/.github/workflows/test-node.js.yml
+++ b/.github/workflows/test-node.js.yml
@@ -1,0 +1,28 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: test - Node.js CI
+
+on:
+  push:
+    branches: [ master ]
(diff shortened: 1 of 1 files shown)


brodybits:
I am taking the liberty to merge immediately since I think this is needed for contributions coming in the near future.

karfau:
Just one question: why did you choose `npm install` over `npm ci`?
(My understanding would be that we should always check what is in the `package-lock.json` and `npm install` can change that file if there is something newer in the specified ranges.)

Reference: https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable

brodybits:
> why did you choose `npm install` over `npm ci`?

That was an unfortunate oversight on my part. Some of my own projects are using `yarn.lock` instead, and some others do not commit any lock file; `npm ci` would not work in either of these cases. Please feel free to raise a new PR to fix it, if you like.

PR #66, opened 2020-06-28
fix: Convert all line separators to LF

### LINKED ISSUE #56: Mixed line endings in the repo
There are 18 files that contain CRLF line endings/ line separators/ line breaks in them:

I found this out the hard way by having huge diffs in #51 .

To detect them I downloaded the zip archive of the repo (to circum went changes git does when checking out code), opened the unzipped folder in my IDE and [ran the related inspection](https://www.jetbrains.com/help/idea/running-inspections.html#run-inspections-manually) "Inconsistent line separators" on the whole folder structure:

```
test/3rd-cases/o3xml.js
test/parse/namespace.js
test/parse/node.js
test/parse/big-file-performance.js
test/parse/simple.js
test/parse/parse-element.js
test/parse/locator.js
test/error/xml-error.js
test/error/error.js
test/html/normalize.js
test/dom/attr.js
test/dom/element.js
test/dom/level3.js
test/dom/clone.js
test/dom/index.js
test/test.js
test/index.js
__package__.js
```

Ideally even when people from different OSes contribute the repo should only contain LF line endings,
for the git config option `core.autocrlf` to work properly.

This can be enforced by adding a [`.gitattributes`](https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings#per-repository-settings) and/or [.editorconfig](https://editorconfig.org/) file into the repo.

karfau:
This just converts the files, no enforcement yet

fixes #56


karfau:
@brodybits Do you think it should be part of this PR to enforce LF using one of the methods described in the ticket?
And if so, which one would you prefer?

PS: I just scanned the changes and I didn't find any multiline XML snippets in those files.

PR #67, opened 2020-06-28
fix: Restore `&nbsp;` behavior from v0.1.27

### LINKED ISSUE #57: Regression regarding handling of `&nbsp;` with "(x)html" mimeType

karfau:
Might be considered a breaking change.

fixes #57

### DIFF
--- a/test/html/normalize.vows.js
+++ b/test/html/normalize.vows.js
@@ -13,13 +13,8 @@ wows.describe('html normalizer').addBatch({
     	assert(dom+'', '<div xmlns="http://www.w3.org/1999/xhtml">&lt;123e>&amp;&lt;a<br/></div>');
     	
     	var dom = new DOMParser().parseFromString('<div>&nbsp;&copy;&nbsp&copy</div>','text/html');
-    	assert.skip(
-    		dom+'', '<div xmlns="http://www.w3.org/1999/xhtml">\u00a0\u00a9&amp;nbsp&amp;copy</div>',
-				'see https://github.com/xmldom/xmldom/issues/57'
-			);
-    	assert(dom+'', '<div xmlns="http://www.w3.org/1999/xhtml"> \u00a9&amp;nbsp&amp;copy</div>');
+    	assert(dom+'', '<div xmlns="http://www.w3.org/1999/xhtml">\u00a0\u00a9&amp;nbsp&amp;copy</div>');
 
-    	
     	var dom = new DOMParser().parseFromString('<html xmlns:x="1"><body/></html>','text/html');
     	assert(dom+'', '<html xmlns:x="1" xmlns="http://www.w3.org/1999/xhtml"><body></body></html>');
 	},
--- a/lib/entities.js
+++ b/lib/entities.js
@@ -69,7 +69,7 @@ exports.entityMap = {
        yacute: "ý",
        thorn: "þ",
        yuml: "ÿ",
-       nbsp: " ",
+       nbsp: "\u00a0",
        iexcl: "¡",
        cent: "¢",
        pound: "£",


brodybits:
Thanks. This would definitely be breaking for anything that does not expect `\u00a0` characters for input with `&nbsp` characters.

karfau:
I think it's a question of definition.
This was the behavior of `0.1.27` changed in `0.2.0 - 0.3.0`(so not a semver issue since major is still `0`),
Now it "breaks" it "back" so I guess we could call it `0.4.0`.
or if you say it was an unintended regression we could include it in `0.3.1`.

Your decision.

test/html/normalize.vows.js
```diff
@@ -13,13 +13,8 @@ wows.describe('html normalizer').addBatch({
     	assert(dom+'', '<div xmlns="http://www.w3.org/1999/xhtml">&lt;123e>&amp;&lt;a<br/></div>');
     	
     	var dom = new DOMParser().parseFromString('<div>&nbsp;&copy;&nbsp&copy</div>','text/html');
```
brodybits, line 15:
```suggestion
        var dom = new DOMParser().parseFromString('<div>&copy;&nbsp;&copy;</div>>','text/html');
```

(simplify the test string and fix the indentation)

Or do we need to keep the "&nbsp&copy" part to test something? If so, I would kinda favor separate test strings, with separate test assertions, to check the `&nbsp;` behavior and to test with "&nbsp&copy".

I do think it would be better to break this test script into smaller test cases, and potentially multiple files, at some point.

TODO: The assertion would have to be updated as well. Unfortunately the GitHub suggested changes feature does not work for multiple lines if some lines were deleted in the middle.

test/html/normalize.vows.js
```diff
@@ -13,13 +13,8 @@ wows.describe('html normalizer').addBatch({
...
-    	assert.skip(
-    		dom+'', '<div xmlns="http://www.w3.org/1999/xhtml">\u00a0\u00a9&amp;nbsp&amp;copy</div>',
-				'see https://github.com/xmldom/xmldom/issues/57'
-			);
-    	assert(dom+'', '<div xmlns="http://www.w3.org/1999/xhtml"> \u00a9&amp;nbsp&amp;copy</div>');
+    	assert(dom+'', '<div xmlns="http://www.w3.org/1999/xhtml">\u00a0\u00a9&amp;nbsp&amp;copy</div>');
 
-
```
brodybits, line 22: 👍

brodybits:
I went ahead with the merge, and the change does now show up on the Stryker dashboard. Thanks!

test/html/normalize.vows.js
karfau: I agree that the tests are not ideal, that's why I raised #72 to discuss the options.

I would avoid to mix this into other PRs to not loose track.

Issue #60, opened 2020-06-26
Add Stryker with Stryker dashboard


brodybits:
__updated:__

~~I would like to do this after finishing review & merge of PR #59.~~

- add Stryker - PR #70
- integrate with dashboard - workaround needed due issue described in <https://github.com/stryker-mutator/stryker-dashboard/issues/156> - PR #77
- add the Stryker badge with link to the Stryker dashboard (PR #77):
  [![Mutation report](https://camo.githubusercontent.com/ee312c4ebce7784ce9f785757eba5d6e33e6d950/68747470733a2f2f696d672e736869656c64732e696f2f656e64706f696e743f7374796c653d666c61742675726c3d687474707325334125324625324662616467652d6170692e737472796b65722d6d757461746f722e696f2532466769746875622e636f6d25324662726f647962697473253246786d6c646f6d2532466d6173746572)](https://dashboard.stryker-mutator.io/reports/github.com/brodybits/xmldom/master)

PR #70, opened 2020-06-29
add Stryker Mutator

brodybits:
__updated:__

https://stryker-mutator.io/ - better than traditional code coverage (I think), as discussed in #60

shows the following summary at this point:

```
---------------|---------|----------|-----------|------------|----------|---------|
File           | % score | # killed | # timeout | # survived | # no cov | # error |
---------------|---------|----------|-----------|------------|----------|---------|
All files      |   59.53 |     1008 |       179 |        807 |        0 |       0 |
 dom-parser.js |   73.04 |      141 |         8 |         55 |        0 |       0 |
 dom.js        |   65.10 |      521 |        87 |        326 |        0 |       0 |
 entities.js   |    5.28 |        5 |         8 |        233 |        0 |       0 |
 sax.js        |   68.36 |      341 |        76 |        193 |        0 |       0 |
---------------|---------|----------|-----------|------------|----------|---------|
```

Seems to be OK in `lib/dom-parser.js`, pretty bad in `lib/entities.js`.

FUTURE TODO item:

- I would also like to integrate this with the [Stryker dashboard](https://dashboard.stryker-mutator.io/), where it should be possible to scan a copy of the code for surviving mutants missing test coverage.

/cc @karfau

## merging

I think these changes should be merged as a single squash commit upon final review.


.gitignore
```diff
@@ -1,2 +1,5 @@
 /node_modules
 /.proof.out
+
+# stryker temp files
+.stryker-tmp
```
karfau, line 5: In case you don't plan to commit the generated report you need to also ignore `/reports`. (At least this is a folder that was generated when I ran it earlier.)

brodybits: I just pushed some of the changes you had originally made for PR #51, which would probably overlap #73 now.

brodybits:
Here is the badge from a fork I made, which links to the online report:

[Mutation testing badge](https://dashboard.stryker-mutator.io/reports/github.com/xpbrew/xmldom-stryker-fork/stryker-master)

karfau:
I'm very happy that the score is actually higher then [when I first attempted it](https://github.com/xmldom/xmldom/pull/51#issuecomment-647014341).
It means the way I converted the tests in the second go is more robust.

brodybits:
> I'm very happy that the score is actually higher then [when I first attempted it](https://github.com/xmldom/xmldom/pull/51#issuecomment-647014341).

That sure is nice progress, thanks!

FYI I did take the liberty to add newline to end of `stryker.conf.json`, which I think is standard editing practice and always done by vim.

karfau:
Sure, I'm not sure why it wasn't there.
I never touched the file, so I guess this i the way that stryker generated it on `init`.

PR #79, opened 2020-07-01
Fixes #75 Add note about import

### LINKED ISSUE #75: Add support for named exports
The current docs indicate that to use the package you need to do:

```js
const DOMParser = require("xmldom").DOMParser;
```

It would be good to have named exports, so that this could be done:

```js
const { DOMParser } = require('xmldom');
```

or

```js
import { DOMParser } from 'xmldom'
```

This would not replace the current approach, but would be an addition allowing following the style of packaging that is being adopted.

ajmas:
Resolves #75

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -32,6 +32,13 @@ var nsAttr = doc.documentElement.getAttributeNS('./lite','x')
 console.info(nsAttr)
 console.info(doc)
 ```
+
+Note: in Typescript and ES6 you can use the import approach, as follows:
+
+```javascript
+import { DOMParser } from 'xmldom'
+```
+
 API Reference
 =====
 


readme.md
```diff
@@ -32,6 +32,13 @@ var nsAttr = doc.documentElement.getAttributeNS('./lite','x')
 console.info(nsAttr)
 console.info(doc)
 ```
+
+Note: in Typescript and ES6 you can use the import approach, as follows:
```
brodybits, line 36:
```suggestion
Note: a more modern import approach is possible in Typescript and ES6, as follows:
```

ajmas: I can make the changes. Had just been trained by other projects with the 'fixes' prefix to allow auto closing of tickets, but I understand each project is different. I'll need to make a new PR to remove that and close this one.

brodybits: I went ahead with the merge, just moved your fixes note from the subject into the comments in the commit message. Your note is shorter so I just kept it as-is. Thanks again!

PR #81, opened 2020-07-01
modernize & improve the example in readme.md

brodybits:
- used `prettier-standard --lint` to reformat the example, with semicolons removed, etc., but with 4-space indentation restored (by hand)
- replace `var` with `const`
- use destructured `const` to import the `DOMParser` object
- add blank lines
- move a `console` logging statement

triggered by discussion in issue #75

related to: PR #79 (add note about ES6 import)

✅ **tested** in a new npm package project in my work area

/cc @ajmas

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -19,19 +19,22 @@ Install:
 Example:
 ====
 ```javascript
-var DOMParser = require('xmldom').DOMParser;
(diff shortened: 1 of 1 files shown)


ajmas:
Looks good to me.

Issue #83, opened 2020-07-03
Duplicate inclusion of entities causes execution failure in Rhino


joestringer:
Observed when including xmldom 0.3.0 in a Rhino 1.7.12 environment:

```
Property "auml" already defined in this object literal.
```

This is a regression caused by ef20262e39a9a188e0f5dc46bc959f4633397706 which mistakenly added duplicates for "auml", "ouml", "aring".

PR #84, opened 2020-07-03
Revert "Add lowercase of åäö in entityMap"

joestringer:
This reverts commit ef20262e39a9a188e0f5dc46bc959f4633397706.

These characters were already included lower in the file in the sorted
order for lower-case runes.

Including duplicates like this causes issues in Rhino 1.7.12:

    Property "auml" already defined in this object literal.

Fixes: #83


brodybits:
Thanks! I may need some time to double-check before merging.

I suspect that the issue could have been avoided had we started using using eslint or a derivative such as standard, as discussed in some other issues.

brodybits:
> This reverts commit [ef20262](https://github.com/xmldom/xmldom/commit/ef20262e39a9a188e0f5dc46bc959f4633397706).

(introduced by PR #23)

I have some nits coming up.

lib/entities.js
```diff
@@ -244,4 +241,4 @@ exports.entityMap = {
        hearts: "♥",
        diams: "♦"
 };
-//for(var  n in exports.entityMap){console.log(exports.entityMap[n].charCodeAt())}
+//for(var  n in exports.entityMap){console.log(exports.entityMap[n].charCodeAt())}
```
brodybits, line 244: This change which reverts some formatting cleanup is something I don't really like about reverts in general.

This line looks pretty useless. I just raised PR #85 to completely remove it.

joestringer: Looks good, how do you want me to handle this PR?

joestringer: Easy way to resolve the conflict will be for me to drop this hunk, which I just did.

brodybits:
Thanks for the contribution. I do have a few more nits for the future:

- The test coverage of `lib/entities.js` is very spotty. I just raised PR #86 to start testing the affected entities, to gain some extra confidence, and it does look good with this change.
- I noticed that the change was contributed from the master branch of the fork. It is generally recommended to contribute changes from another, temporary branch instead, to avoid likely divergence. PRs are generally merged as squash commits in this project.

As I had implied before, it is not very nice when a revert would undo formatting cleanup. But I think taking it out would make it not a truly complete revert. I think PR #85 solved it by simply getting rid of the useless line.

Once we have better test coverage, Stryker should show us if we have duplicated string entities again.

On a positive note: kudos for putting the information into the revert commit itself. This should make it easier for people to understand the rationale without having to dig through GitHub.

Thanks again for the contribution!

PR #86, opened 2020-07-03
test European HTML entities

brodybits:
The immediate purpose is to add test coverage for the entityMap entries that are affected by PR #84. This should give us some extra confidence that PR #84 really does not break anything and increase the mutation score a very little bit.

It would be nice to continue adding test coverage for more European and other international entities in `lib/entities.js`, which should increase the mutation score even further, someday.

P.S. The downside is that this would become one more thing to be reworked when moving from vows to something more current such as Jest.


test/html/normalize.vows.js
```diff
@@ -85,5 +85,11 @@ wows.describe('html normalizer').addBatch({
     	
     	var dom = new DOMParser().parseFromString('<textarea>alert(a<b&&c?"<br>":">>");</textarea>','text/xml');
     	assert(dom+'', '<textarea>alert(a&lt;b&amp;&amp;c?"<br/>":">>");</textarea>');
+	},
+    'Euro characters': function () {
+        var dom = new DOMParser().parseFromString('<div>&Auml;&auml;&Aring;&aring;&AElig;&Ouml;&ouml;&Oslash;&euro;</div>','text/html');
+        // For the future, it may be nicer to use \uxxxx in the assert strings
+        // rather than pasting in multi-byte UTF-8 Unicode characters
+        assert(dom+'', '<div xmlns="http://www.w3.org/1999/xhtml">ÄäÅåÆÖöØ€</div>');
```
karfau, line 93: Just completed the list.
```suggestion
        var dom = new DOMParser().parseFromString('<div>&Auml;&auml;&Aring;&aring;&AElig;&aelig;&Ouml;&ouml;&Oslash;&oslash;&Uuml;&uuml;&euro;</div>','text/html');
        // For the future, it may be nicer to use \uxxxx in the assert strings
        // rather than pasting in multi-byte UTF-8 Unicode characters
        assert(dom+'', '<div xmlns="http://www.w3.org/1999/xhtml">ÄäÅåÆæÖöØøÜü€</div>');
```

test/html/normalize.vows.js
```diff
@@ -85,5 +85,11 @@ wows.describe('html normalizer').addBatch({
     	
     	var dom = new DOMParser().parseFromString('<textarea>alert(a<b&&c?"<br>":">>");</textarea>','text/xml');
     	assert(dom+'', '<textarea>alert(a&lt;b&amp;&amp;c?"<br/>":">>");</textarea>');
+	},
+    'Euro characters': function () {
```
brodybits, line 89:
```suggestion
    'European entities': function () {
```

I don't know if it is worth making this change or not:)

brodybits:
Unfortunately this merge did not go right for me.

I meant to merge this in a single squash commit, which is what I normally do, but it ended up making a merge commit which leaves a little bit of a mess in the git commit history. I did see a message that the head had changed, with a button to try again. After I clicked the button to try again then it did merge, in a merge commit. Argh!!

Issue #72, opened 2020-06-29
Improve test coverage with future-proof testing strategy


karfau:
We just restored the existing `vows` test suite.

But I don't think we should add new tests to it (or only in a limited manner).

- there are many tests with multiple assertions, first failing hides other potential failling ones
- `vows` is another abandoned test framework
- generating those kind of repetitive tests is more verbose then in current frameworks (like jest)

I think this is again a high prio issue since it lower the barrier to contribution **and** we will hopefully add a lot of tests for all the known issues. (I imagine to add branches or even draft PRs with a failing test case as a first step of every investigation.)

Suggested LOD(level of done):
- [x] decide for a test framework that is maintained and well known to make contributions easy: we decided using jest, it landed with #114 
- [x] add a first (green, but make sure it can fail) test to the repo including CI on github actions => also part of #114 
- [ ] propose a concept how to structure the tests in a consistent manner (e.g. a folder per module, test suite per function or "topic"), see
  - https://github.com/xmldom/xmldom/issues/72#issuecomment-678482473
  - https://github.com/xmldom/xmldom/issues/72#issuecomment-682345823
  - [ ] document and apply the proposal once it's agreed upon
- [ ] cover more parts of the DOM API:
  - [ ] [`setAttributeNode` and `setAttributeNodeNS`](https://github.com/xmldom/xmldom/issues/72#issuecomment-761727690)
- [ ] Improve Stryker score

@ owners & authors Please keep this description up to date with results from the discussion below.

brodybits:
I took the liberty to pin this issue and add a dash, which should hopefully make it extra clear to maintainers like myself that it is proof as in future-proof and not proof as in the recently removed test tool.

I suspect it should be straightforward to port one script at a time from vows to the Jest format with the `describe`, `it`, and `expect` functions. Most scripts under 100 lines, largest script just under 150 lines:

```sh
brodybits@brodybits-mini-macbook xmldom % wc -l test/**/*.vows.js
     146 test/assert.vows.js
      65 test/dom/attr.vows.js
      24 test/dom/clone.vows.js
     148 test/dom/element.vows.js
      15 test/dom/fragment.vows.js
      48 test/dom/ns-test.vows.js
      15 test/dom/serializer.vows.js
      95 test/error/error.vows.js
      71 test/error/xml-error.vows.js
      94 test/html/normalize.vows.js
      18 test/parse/file-test.vows.js
      76 test/parse/locator.vows.js
[...]

Issue #29, opened 2020-03-04
standard(ize) source code formatting


brodybits:
from review of pr #28 I discovered that code is sometimes indented with only tabs and sometimes with a combination of tabs and spaces. Also at least one occasion of code with no semicolon.

Most drastic would be to use `standard`, `semistandard`, or `prettier-standard` to reformat the code in what I think is a nicer style. (`prettier-standard` actually uses the `prettierx` fork of Prettier that is maintained by myself, which could be seen as an advantage or perhaps even a disadvantage)

Less drastic would be to apply some simplistic rules such as only indent with tabs and always use semicolons to achieve better consistency with minimal code churn. I think it should be possible to do this with custom eslint rule configuration.

kethinov:
I personally use `standard` on my other projects. I almost threw it into that PR but it would've been an epic diff.

awwright:
I'd like to submit a PR for an `.eslintrc.js` file, and an `.editorconfig` file, but it'd probably be a good idea to figure some things out first...

Since there doesn't really seem to be a standard, it might be a good idea to not convert everything all at once... I'm playing around with a PR for a bare-minimum `.eslintrc.js` file for obvious problems; and then you can add code style preferences later.

For indents, my vote would be to keep tabs. Not to rehash the tabs v. spaces holy war or anything, but I like customizing my indent size, and I like it when you can't accidentally highlight half an indent (as an example: lib/entities.js is indented 7 spaces).

Thoughts?

PR #89, opened 2020-07-13
Port custom assertions to mostly strict assertions

awwright:
I'd like to start chipping away at converting the test suite, I've converted a Vows suite to Mocha before. This one looks pretty straightforward, actually.

I think the first logical step here is to start with the assertions themselves, and use the Node.js core library and only the builtin functions.

In particular, `assert.isTrue` is a monkey-patch by Vows. I found most all the tests can use `assert.strictEqual`. There's a few that test if an object is cast to a string, these use `assert.equal`. There's two more that I think don't meet the DOM standard if you use `===`, these use `assert.equal` and are marked with a FIXME.


karfau:
Thx for the time invested.

From the feedback I got in my first attmept in #51 to do something similar I learned that the owners prefer to keep those tests as they are.
I also wrote the "internal" `assert` to be able to switch between `assert.isTrue` and `assert.isEqual` using the environment variable `XMLDOM_ASSERT_STRICT=1`. Just removing it again feels strange to me.

I also opened #72 to discuss how to handle testing in the future.
Due to the current status of the code base, my personal suggestion would be to not touch the vows tests and instead start fresh with a new test suite that is better structured.
There has not been a decision yet what test framework to use, but jest and mocha are of course reasonable approaches.

But since I'm only a contributor, I hope the owners will also add their perspective to the mix.

.editorconfig
```diff
@@ -0,0 +1,9 @@
+# http://EditorConfig.org
+root = true
+
+[*.js]
+charset = utf-8
+indent_style = tab
+indent_size = 2
+end_of_line = lf
+insert_final_newline = true
```
karfau, line 9: Personally, I believe this is a good idea, but should be added in a separate PR where just this topic can be discussed. See #29

awwright: Oops, yes of course

awwright:
@karfau I don't think having the the tests be dependent on the setting of a variable is a very good idea... I would assume tests cover every situation by default (or at at least as much as the execution environment can do).

But also for the vast majority of the tests, I don't think that makes any sense. If you're testing the length of an array, or the contents of a string, there's no reason to do that non-strict.

Which assertions specifically need the old behavior?

karfau:
@awwright did you read the conversation in #51 (and potentially follow up #59 ) that lead to the creation of those assertions? I'm not really willing to repeat the arguments stated there (, also since they are not mine).

I agree that there are many tests that are not very useful, redundant or hard to maintain.
But's it's what we have right now.

But I would rather invest into a properly structured test suite using modern tooling, that makes the vow's test suite obsolete (by "covering all it covers" and more), instead of "tweaking" the vows test suite and risking that we cover less afterwards. But as I said, it's just my personal preference.
I'm not part of the organization making decisions in this repo.

brodybits:
Hi I am reviewing the changes right now.

I would actually favor converting the existing custom assertions to strict assertions, assuming that it does not affect testing speed too aversely. The testing speed is pretty important for the Stryker mutation testing that is already in place.

I would also not mind if we would port the existing tests from vows to something more universally accepted such as Jest. (I would actually not favor Mocha, but that is personal preference).

I would definitely agree with @karfau that the existing tests are pretty bad, and we may need to replace them all over time, as discussed in #72.

When the custom assert function was introduced in PR #59, my understanding was that this was only a temporary solution to restore the existing test coverage.

I did also take the liberty to update the title.

In short, I do think this is a step in the right direction, even if it is not where we really want to be in the end.

test/parse/locator.vows.js
```diff
@@ -71,6 +73,6 @@ var xml= [
     var doc = parser.parseFromString('<root>\n\t<err</root>', 'text/html');
     var root = doc.documentElement;
     var textNode = root.firstChild;
-	assert.isTrue(/\n@#\[line\:2,col\:2\]/.test(error.join(' ')),'line,col must record:'+JSON.stringify(error));
+	assert(/\n@#\[line\:2,col\:2\]/.test(error.join(' ')),'line,col must record:'+JSON.stringify(error));
```
brodybits, line 76: A bit hard to read & understand, how about this idea:

```suggestion
	assert(/\n@#\[line\:2,col\:2\]/.test(error.join(' ')), 'line,col must record: ' + JSON.stringify(error));
```

or maybe even this:

```suggestion
	assert(/\n@#\[line\:2,col\:2\]/.test(error.join(' ')), `line,col must record: ${JSON.stringify(error)}`);
```

or one step further, like what I think Prettier would do:

```suggestion
        assert(
          /\n@#\[line\:2,col\:2\]/.test(error.join(' ')),
          `line,col must record: ${JSON.stringify(error)}`
        );
```

test/parse/namespace.vows.js
```diff
@@ -1,33 +1,37 @@
...
        var root = dom.documentElement;
-       assert(root.namespaceURI, 'http://test.com')
-       assert(root.lookupNamespaceURI(''), 'http://test.com')
-       assert(root.firstChild.namespaceURI, 'http://test.com')
-       assert(root.firstChild.lookupNamespaceURI(''), 'http://test.com')
-       assert(root.firstChild.getAttributeNode('attr').namespaceURI, null)
+       assert.strictEqual(root.namespaceURI, 'http://test.com');
+       assert.strictEqual(root.lookupNamespaceURI(''), 'http://test.com');
+       assert.strictEqual(root.firstChild.namespaceURI, 'http://test.com');
+			 assert.strictEqual(root.firstChild.lookupNamespaceURI(''), 'http://test.com');
```
brodybits, line 15: fix alignment of this line?

test/parse/test-doc-whitespace.vows.js
```diff
@@ -9,16 +12,16 @@ wows.describe('errorHandle').addBatch({
...
 	},
 	'test':function(){
 		var description = "<p>populaciji (< 0.1%), te se</p>";
 		var doc = new DOMParser().parseFromString(description, 'text/html');
-		assert(doc.toString(), '<p xmlns="http://www.w3.org/1999/xhtml">populaciji (&lt; 0.1%), te se</p>')
-	}
-}).export(module)
+		assert.strictEqual(doc.toString(), '<p xmlns="http://www.w3.org/1999/xhtml">populaciji (&lt; 0.1%), te se</p>');
+	},
+}).export(module);
```
brodybits, line 24: I think these should be reverted for now:

```suggestion
	}
}).export(module)
```

test/dom/attr.vows.js
```diff
@@ -28,30 +30,30 @@ wows.describe('XML attrs').addBatch({
...
     },
     "override attribute":function(){
     	var root = new DOMParser().parseFromString("<xml xmlns:a='a' xmlns:b='b' xmlns='e'><child/></xml>",'text/xml').documentElement;
     	root.setAttributeNS('a','a:a','1');
-    	assert(root.attributes.length, 4);
+    	assert.strictEqual(root.attributes.length, 4);
 //not standart
 //    	root.firstChild.setAttributeNode(root.attributes[0]);
-//    	assert(root.attributes.length, 0);
+//    	assert.strictEqual(root.attributes.length, 0);
```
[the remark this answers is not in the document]
awwright: Comment them, eventually? Idk what exactly, but if they do get uncommented, this will definitely want to use strictEqual.

lib/dom-parser.js
```diff
@@ -1,3 +1,5 @@
+"use strict";
+
```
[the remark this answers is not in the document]
awwright: This snuck in by mistake

awwright:
@brodybits Wrt the rest of the review... It seems to me maybe we should do a PR to normalize the whitespace first, then I can apply these fixes after that. Thoughts?

brodybits:
> do a PR to normalize the whitespace first

Yes. I think it would be easier to get the lint in place first. Thanks.

awwright:
I rebased over the updates from #121.

brodybits:
I have updated the title, and I pushed a bunch of updates to give us a green build and make some improvements.

I can think of a couple followup items:

- finish removing use of Vows, which is now done in draft PR #126
- use standard Jest expect statements, which would more likely show multiple failed expectations in each test case

karfau:
> * use standard Jest expect statements, which would more likely show multiple failed expectations in each test case

@awwright Are you still investing time to porting the assertions to strict ones?
I agree that we should rather move them to `expect` assertions provided by jest in one go.

Or did I misunderstand you @brodybits ?

brodybits:
I could go either way. I would not mind a multi-step approach like this:

- port assertions to strict assertions - thanks to @awwright for your efforts!
- we use Prettier to reformat, like I proposed in PR #123
- then we can convert the strict Node.js assertions to `expect` assertions

awwright:
I was proposing that, after the code is formatted, to first port the assertion calls (as in this PR). That would make changing over the test runner a minimal amount of effort. Then working on fixing tests and adding coverage.

...it seems the test runner has mostly been converted, though.

I favor using `assert`. It's simple and straightforward.

brodybits:
@awwright we have already switched from Vows to Jest. There are still a few more steps to completely remove Vows, which I have already done in PR #126.

brodybits:
Another thing is that I had spent several hours getting this PR to run with a green build, and find it very unfortunate to see my work thrown away here. (I do have my work saved in my own fork though.)

At this point I am thinking to go ahead with PR #123 with the Prettier reformatting and then convert the custom assertions directly into Jest-style `expect` assertions.

awwright:
@brodybits I'm sorry, when was that? Was this PR introducing build errors?

brodybits (changes requested):
I think it would be easiest for us to apply the Prettier formatting (PR #123) and then port the custom assertions directly to Jest `expect` statements. So there should be no need to keep working on this PR.

Thanks to @awwright for the efforts.

awwright:
@brodybits Well, OK, but I'm trying to figure out what changes you're talking about, do you mean a PR? I'm curious which errors you're referring to.

brodybits:
I had pushed the changes to your `convert-assertions` branch on GitHub. See this: https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/committing-changes-to-a-pull-request-branch-created-from-a-fork

Unfortunately I don't have much time to explain the errors that I had fixed, due to some other priorities.

In any case, your contributions did help us get the code cleaner. You should be able to see that you have co-authorship credit on some recent changes that we made.

awwright:
I mean, I'm not sure what else to say... GitHub doesn't show any history, but it shows my rebases and commits that I pushed, and I see you mentioning this from other PRs, but it doesn't show any commits pushed to my branch.

PR #90, opened 2020-07-14
Lint with tabs & enable strict mode

awwright:
This adds an `.eslintrc.js`, `.editorconfig`, and normalizes the whitespace to 1-tab indents.

Btw, whitespace can be ignored from GitHub by adding `?w=1` to a diff URL: <https://github.com/xmldom/xmldom/pull/90/commits/2bbb7e2459abc64f7270f5f08eab8d8022137f96?w=1>

This is mostly just a suggestion, it's easy enough to go with a different style if something else is more popular.

After this:

- [ ] Fix remaining eslint errors with switch/break statements
- [ ] Add eslint to CI tests
- [ ] Enable additional eslint tests


brodybits:
Thanks. I have mixed feelings about using tabs at this point. My biggest worry with tabs is that they do not look so nice on Unix editors, especially on Linux & Unix.

I am starting to see your point in #29 that tabs can help avoid issues like 7-space indentation and 2-space vs 4-space indentation. It is also nice to see not all lines of the code suddenly reindented with spaces. I would be interested in what the other contributors & maintainers have to say about this idea.

It is possible to configure eslint in a YML file, which may be a little easier for us to update over time. It would also be possible to configure eslint per subdirectory tree, which would make it possible to format `lib` and `test` in separate PRs. An idea might be to format the tests first, then continue fixing the assertions in #89 before cleaning up `lib`.

I am also thinking we could use Prettier at the same time to solve both indentation and code formatting at the same time. I think this would be better than the extra code churn that would come from doing this in multiple steps. I think it should be pretty straightforward to add eslint and Prettier together using eslint-plugin-prettier.

A final comment on lib is that the test coverage of `lib/entities.js` is pretty bad, as you can see by following the report from the mutation score badge. It might be nice to cover remaining entries in `lib/entities.js` before reformatting it.

awwright:
Thanks, I appreciate the consideration. Also note:

- GitHub reads `.editorconfig`, and will set tab widths to whatever is specified there.
- On Linux and macOS you can run `tabs -4` (or put it in your shell rc) which sets the tab stops to every 4 columns. (or whatever other number you like).

brodybits:
That makes sense for me, and thanks for the tip to use `?w=1`. I took the liberty to update the title, would like to give others some time for feedback before continuing with a possible merge.

Do you think my idea to use eslint-plugin-prettier to clean up the formatting in one shot makes sense?

awwright:
@brodybits I don't have any experience with that, but doing more rather than less makes sense to me. I've usually found `--fix` to be sufficient for formatting, except for some mixed-spaces-tabs problems.

brodybits:
I saw an excellent argument recently on Twitter: accessibility

I think I lost the first one I saw, think I found another one:

<blockquote class="twitter-tweet"><p lang="en" dir="ltr">I was firmly in the camp of spaces only in source code (entered with the tab key, of course, and proper editor settings) until I read this about the accessibility benefits of using tab characters for indentation:<a href="https://t.co/TIASFjXcL8">https://t.co/TIASFjXcL8</a></p>&mdash; Tommy Williams (@twwilliams) <a href="https://twitter.com/twwilliams/status/1220792888235134976?ref_src=twsrc%5Etfw">January 24, 2020</a></blockquote>

https://www.reddit.com/r/javascript/comments/c8drjo/nobody_talks_about_the_real_reason_to_use_tabs/

lib/dom.js
```diff
@@ -1,3 +1,5 @@
+"use strict";
```
karfau, line 1: I don't have a clear understanding of the impact, might this be a breaking change?
According to MDN "[Strict mode changes both syntax and runtime behavior.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#Changes_in_strict_mode)".
So I would rather wait for more tests to be available before making such a change.

lib/dom.js
```diff
@@ -1188,6 +1191,23 @@ function cloneNode(doc,node,deep){
...
+			if(node.nodeType!==7 && node.nodeType !==8){
+				buf.push(getTextContent(node));
+			}
+			node = node.nextSibling;
+		}
+		return buf.join('');
+	default:
+		return node.nodeValue;
+	}
+}
```
karfau, line 1210: What is the reason for moving this function here? At first glance is seems to be unrelated to the linting changes. If so you could extract it into a separate PR.

lib/dom.js
awwright: It's possible to statically determine all of the new errors strict mode might throw. Iirc, eslint raises errors for all of these cases by default.

lib/dom.js
awwright: This function was previously inside an "if" statement, which is ambiguous behavior because of function hoisting; this is prohibited in strict mode.

karfau:
Maybe we can already land the `.editorconfig` in an extra PR? So that all new contributions don't need to change when this PR lands?

brodybits:
> Maybe we can already land the .editorconfig in an extra PR? So that all
> new contributions don't need to change when this PR lands?
>

+1 I will likely do that soon

>
> --
Sent from my mobile

.editorconfig
```diff
@@ -0,0 +1,9 @@
+# http://EditorConfig.org
+root = true
+
+[*.js]
+charset = utf-8
+indent_style = tab
+indent_size = 2
```
brodybits, line 7: I think `indent_size` is not needed with `indent_style = tab` but am not 100% sure. @karfau?

awwright: It does have the effect of setting the size when previewed on GitHub, at the very least.

brodybits: It stays then, thanks.

brodybits:
I have started making these changes in the following PRs:

- #104 - add `.editorconfig` now merged ✅
- #102 - eslint with tabs & remove trailing whitespace
- #110 - eslint with Prettier formatting

As I said in PR #110, I would now favor that we do the formatting with Prettier all at once.

I think the changes to enable strict mode are also needed and should be done after we have finished with the eslint and/or Prettier formatting.

PR #91, opened 2020-07-15
remove ES6 syntax from getElementsByClassName

rg1:
The new method getElementsByClassName uses ES6 syntax. This is a problem when trying to polyfill an older engine.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -609,10 +609,10 @@ Document.prototype = {
 	},
 	
 	getElementsByClassName: function(className) {
-		const pattern = new RegExp(`(^|\\s)${className}(\\s|$)`);
-		return new LiveNodeList(this, base => {
+		var pattern = new RegExp("(^|\\s)" + className + "(\\s|$)");
+		return new LiveNodeList(this, function(base) {
 			var ls = [];
-			_visitNode(base.documentElement, node => {
+			_visitNode(base.documentElement, function(node) {
 				if(node !== base && node.nodeType == ELEMENT_NODE) {
 					if(pattern.test(node.getAttribute('class'))) {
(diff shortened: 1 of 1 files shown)


brodybits:
Thanks. Can you give us an idea what kind of engine or browser you need this for?

In my own experience, these changes could help on Android 4.4, which is extremely old and losing support from toolkits such as Apache Cordova.

If we do accept this proposal, it would be nice to configure eslint to reject the ES6 syntax (maybe part of PR #90).

/cc @karfau @kethinov @awwright

awwright:
I was just going to chime in: if we standardize on supported platforms, this should include an eslint rule for it.

If this is the only thing stopping use in a handful of platforms, +1. I'm not worried about syntax sugar, I'm more worried about getting to use optimizations like `Set` and `Map`.

karfau:
With the massive amount of usage out there, introducing a syntax that is not supported is a breaking change. So we should really define what platforms to support (and for how long).

I think this PR is good as a first step (didn't check the code details, just talking about the intention here). Next one should be to 
1. define the supported platforms
2. take measures to prevent future breaking changes by one or multiple of
   - run tests on them if possible?
   - offer a polyfilled version?
   - eslint rules to prevent unsupported syntax (not sure how much it can cover)

rg1:
The platform in question is [Jint](https://github.com/sebastienros/jint). This is the only place that is currently using modern syntax and with my change, I can use xmldom even in the old/stable branch.

For the record, the other platform I know of that won't support this syntax is IE11.

xqdoo00o:
I have the IE11 error too, and fixed as this commit said. It's a little weird only this function but no other funcs use es6 syntax.

PR #95, opened 2020-07-21
configure Dependabot

brodybits:
(npm only)

https://docs.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates#example-dependabotyml-file

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+
+updates:
+  - package-ecosystem: "npm"
+    # Look for `package.json` and `lock` files in the `root` directory
+    directory: "/"
+    # Check the npm registry for updates every day (weekdays)
+    schedule:
+      interval: "daily"


brodybits:
Thanks @karfau. Renovate definitely seems to be more flexible, and (much) less noisy, so I am using it on some of my own projects including:

- https://github.com/brodybits/prettierx
- https://github.com/brodybits/create-react-native-module

What triggered me to propose Dependabot is that it already took the initiative to propose a minor transitive dev security update (PR #94) and seems to be the most widely used. I think another thing is that Renovate is only free for individual projects.

You are right about the default open PRs limit, and I would generally favor increasing it. But this project has no production dependencies, only 5 dev dependencies at this point. And I think the test run time is pretty fast on PRs (Stryker is only run on pushes to the master branch).

That said, I would favor updating transitive dependencies such as lodash by regenerating the package lock file.

karfau:
Just to state the potentially obvious: having security related updates from dependabot and others from renovate is totally possible and works seamlessly in my repos.

I was not aware that it's not free for open source repos from organizations. This really surprises me.

As I said, I think it's a first step into the right direction, so no objections from my side, just comments.

brodybits:
I was mistaken about Renovate, looks like it is free for open-source orgs but probably not for private org projects: https://github.com/marketplace/renovate

And you are right about using both bots, though I generally like to keep just one in control if possible.

My final argument is that for such a low number of dependencies, we may as well favor the "built-in" bot over adding a third-party application. Resting my case now:)

karfau:
:+1:

brodybits:
I think Renovate would really shine on updating the Stryker-Mutator monorepo by updating all packages in a single PR, while Dependabot would make the updates in multiple PRs.

brodybits:
Closing in favor of PR #108 to configure Renovate bot

PR #102, opened 2020-08-17
eslint with tabs; remove trailing spaces

brodybits:
as discussed in PR #90, using `npx eslint --init` to generate `.eslintrc.yml` as a starting point

NOTES (included in one of the commits):

- `git diff -b` / `git show -b` should show very limited output resulting from eslint --fix with these changes
- adding `?w=1` to URL of the commit or PR files link on GitHub should hide most of the output from the whitespace changes

TODO item(s):

- [ ] I think we should improve the mutation testing score before continuing with changes like these. While whitespace changes should not be a major risk, I would feel much safer about improving the test coverage before making this kind of change.

Some followup items:

- consider using Prettier or prettierx, which should give us more consistent formatting
- enable `extends: 'eslint:recommended'` then remove `no-trailing-spaces: error`
- consistent quotes & semicolon style
- cleanup some areas of code affected by these changes


brodybits:

PR #105, opened 2020-08-17
Reduce time needed for stryker

### LINKED ISSUE #82: Improve stryker execution time regarding test command
as discussed in #77 after merging

- [x] improve package.json `scripts` structure => #87 
- [ ] test improved performance in github actions

I will focus on these tasks before any other.

karfau:
By not running the tests for test/assert.js I think we can reduce the duration of the stryker run.
They shouldn't change the result since they are only test helpers that don't change how the actual code is tested.

- [ ] needs to be verified by running it (on github?)

Fixes #82

### DIFF
--- a/stryker.conf.json
+++ b/stryker.conf.json
@@ -14,5 +14,8 @@
     "project": "github.com/brodybits/xmldom",
     "version": "master"
   },
-  "coverageAnalysis": "all"
+  "coverageAnalysis": "all",
+  "commandRunner": {
+    "command": "npm run test:unit"
+  }
 }


brodybits:
I think we need to update `@stryker-mutator` packages to 3.3.1 because of <https://github.com/stryker-mutator/stryker/issues/2286>.

I can do this manually or we can set up the Renovate bot to do the update for us.

I think this is good, though not yet convinced it will save a really huge amount of mutation testing time. I think the upcoming Stryker mutation switching feature will save way more time than this one. But I may be proven wrong here:)

brodybits:
I tried `npm run stryker` with `@stryker-mutator` packages updated to `3.3.1`. It does seem to run a *little* faster. (I tried it on a private [16 vCPU] cloud instance which runs way faster than my little macBook.).

I think this is good to go now that Stryker update was done by PR #109.

brodybits:
Thanks @karfau, I just edited & cleaned up the commit message a little.

PR #106, opened 2020-08-17
add & enable eslint

brodybits:
mostly extracted from PR #102; one or more upcoming PRs will be based off of this one

~~It is an open question whether this PR should be merged or be superseded by another eslint PR.~~

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,11 +1,10 @@
 env:
   browser: true
   commonjs: true
-  es2020: true
   node: true
 extends: 'eslint:recommended'
 parserOptions:
-  ecmaVersion: 11
+  ecmaVersion: 6
 rules:
   indent:
     - error
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,0 +1,21 @@
+env:
+  browser: true
+  commonjs: true
+  es2020: true
+  node: true
+extends: 'eslint:recommended'
+parserOptions:
+  ecmaVersion: 11
+rules:
+  indent:
+    - error
+    - tab
+  linebreak-style:
+    - error
+    - unix
+  quotes:
+    - error
+    - single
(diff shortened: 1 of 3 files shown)


brodybits:
I would appreciate a quick review of this PR & PR #107.

I would like to merge both of these PRs if there are no criticisms in the next few days. Thanks!

.eslintrc.yml
```diff
@@ -0,0 +1,23 @@
+env:
+  browser: true
+  commonjs: true
+  node: true
+# FUTURE TBD:
+# extends: 'eslint:recommended'
+parserOptions:
+  ecmaVersion: 6
```
karfau, line 8: Question: Is this related to the ES6 syntax topic?

brodybits: Yes and no, good question.

In general I wanted to enable the lowest ES version possible with the existing code base when starting to add eslint. I had wanted to do `ecmaVersion: 5`, it worked with 100% with `lib` but not with `test`, so I went with `ecmaVersion: 6`.

Considering that PR #107 *does* seem to solve the problem with enforcing the older ES version in `lib`, I am wondering if we need to keep an outdated ecmaVersion number here.

karfau: I'm not familiar with configuring eslint, but I think being explicit is often better then relying on implicit things. Less "magic".

So if there is not a good reason to remove the option I would leave it as is.

brodybits: I just put it back to `ecmaVersion: 11` with `es2020: true` in the env options above, according to the output of `npx eslint --init`.

My rationale is that I would not object to ES2020 code in tests, and we need PR #107 with eslint-plugin-es5 to solve the problem with older browsers. But I can go back to `ecmaVersion: 6` again.

karfau: Perfect!

PR #87, opened 2020-07-04
test: Restructure & improve npm test scripts

karfau:
Tests for `test/assert.js` need to run before the vows testsuite, not as part of them.
- Rename `test/assert.vows.js` to `test/assert.vows.checks.js`
- Script `test:assert` only runs those tests
- Script `test:assert:strict` runs them with strict assertions
- Script `test:unit` and `test:vows` are currently the same (until we add another testing framework to the mix)

In the end `npm [run] test` executes the same amount of tests on CI but it's more easy to run a subset.

Added devDependency [`npm-run-all`](https://www.npmjs.com/package/npm-run-all)

https://github.com/xmldom/xmldom/issues/82
https://github.com/xmldom/xmldom/pull/77#issuecomment-652690183

PR #107, opened 2020-08-17
use eslint-plugin-es5 to enforce ES5

brodybits:
The purpose is to avoid ES6 items that may cause issues with older engines, as discussed in PR #87.

This should guard against newer ES6 syntax that could lead to issues with some ES5 environments such as:

- Jint - https://github.com/sebastienros/jint
- Duktape
- IE11
- Android pre-5.0 WebView (already past EOL)

This PR is based on PR #106.

I would appreciate a quick review from someone, if possible.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -3,6 +3,8 @@ env:
   commonjs: true
   es2020: true
   node: true
+plugins:
+  - es5
 # FUTURE TBD:
 # extends: 'eslint:recommended'
 parserOptions:
--- a/lib/.eslintrc.yml
+++ b/lib/.eslintrc.yml
@@ -1,0 +1,2 @@
+extends:
+  - 'plugin:es5/no-es2015'
--- a/package-lock.json
+++ b/package-lock.json
@@ -1008,6 +1008,12 @@
         }
       }
     },
+    "eslint-plugin-es5": {
+      "version": "1.5.0",
+      "resolved": "https://registry.npmjs.org/eslint-plugin-es5/-/eslint-plugin-es5-1.5.0.tgz",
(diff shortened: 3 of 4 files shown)


.eslintrc.yml
```diff
@@ -0,0 +1,25 @@
+env:
+  browser: true
+  commonjs: true
+  node: true
+extends:
+  # FUTURE TBD:
+  # - 'eslint:recommended'
+  - 'plugin:es5/no-es2016'
+parserOptions:
+  ecmaVersion: 6
```
brodybits, line 10: The question <https://github.com/xmldom/xmldom/pull/106/files/2181dcbda3b0e27d9033a6b3dd988a912995d6cd#r471948818> got me thinking if we need these constraints in the top-level `.eslintrc.yml`, now that we have the constraint in `lib/.eslintrc.yml`.

I had actually wanted to enforce `'plugin:es5/no-es2015'` at the top level, and only relax it in `test`, but I don't know how to do this and may raise an issue on the plugin. Any ideas would be awesome.

Thanks for the quick reviews. I will probably merge these later today.

karfau: Well I think configuring it to a more current version on the top level and only restricting it in lib makes a lot of sense, assuming we might be looking into ways to allow more modern syntax also in `lib` at some point in time. In that case we would only need to remove the limitation for `lib`.

brodybits: I have now put ecmaVersion back to 11 and added back es2020: true in env above, like I did in PR #106, removed `'plugin:es5/no-es2016'`, and added explicit plugins with es5 above, hope it is OK. I put this PR in draft since I would like to merge PR #106 first.

Please let me know within the next day if you see any issues, otherwise I would like to finish with merging #106 & #107.

karfau: All very reasonable, Go for it!

Issue #103, opened 2020-08-17
Improve test coverage


brodybits:
I think the mutation testing score is a bit too low to feel safe and confident about the various refactoring and other improvements that are wanted by myself and the user community in general.

It is my understanding from @karfau in issues #72 & #99 that we want to transition to a more modern test framework such as Jest before adding too many more tests, 100% agreed on my part.

I think it would be ideal to improve the test coverage before 0.4.0 release (discussed in #99) but will understand if this is not possible.

I would definitely consider this issue to be a major blocker of future work and will pin it in first place.

PR #110, opened 2020-08-17
eslint with Prettier formatting [with semicolons]

brodybits:
Based on PRs #102, this is a proposal to format the code with Prettier.

Given that this PR overwrites such a large proportion of the code, I would favor that we skip PR #102 and incur these updates all at once.

I would favor that we improve the test coverage (issue #103) before merging these updates.


brodybits:

PR #112, opened 2020-08-21
test: Add jest testsuite with `xmltest` cases

karfau:
The only thing the current test suite does is to run ton's of cases and take a snapshot of the output and errors, so that hopefully a test will fail if we make a breaking change.
(It doesn't cover breaking API changes of course, but sets the basis since it adds jest and any added test will be part of CI testing.)

Added devDependencies:
- https://github.com/karfau/xmltest and it's peerDeps
  - https://npms.io/search?q=yauzl
  - https://npms.io/search?q=get-stream

New helper Module `get-test-parser.js`:
To simplify the very common use case of testing `DOMParser` results.


karfau:
I'm having some trouble with the async setup for reading from the zip file.
I learned that with jest, it's not possible to to have async code in describe so I can not create tests dynamically from data if it's only available via an async call. (`beforeAll` doesn't help since it's executed after all describes.) I will need to provide a synchronous API to the test data somehow, hopefully this will be possible by using `globalSetup` hook in jest that can be async.

test/createParser.js
```diff
@@ -0,0 +1,16 @@
+var {DOMParser} = require('../lib/dom-parser')
+
+function createParser ({errorHandler, errors = {}, locator = {}} = {}) {
+  errorHandler = errorHandler || ((key, msg) => {
+    if (!errors[key]) errors[key] = []
+    errors[key].push(msg)
```
brodybits, line 6: minor nit: I think newly added code should avoid overwriting function parameters unless absolutely necessary, which I think is not the case here.

nit 2: I find the file name `createParser.js` a little too generic. If it is intended to be used by other tests in the future, I would favor calling it something like createTestParser.js. If it is only expected to be used for xmltest, I would favor reflecting that in the module name.

another minor nit: I think all other modules in `lib` and `test` are in kebab case, which I would probably favor.

karfau: In this case it's actually nessesary to pass those parameters, tsince this way the "errors, warnings and fatals" are part of the snapshot assertion.Should one of them go missing or change the test would break.
The error handlers are part of the public API and the code to handle them the same way in multiple tests is quite verbose, to I decided to extract them.
Since it's still work in progress, I really don't bother to much with names.
I'm also not sure if we should use it outside of the `xmltest` "suite", so I'm fine with whatever name you prefer.

> kebab case

You are right! I really didn't realize that "convention" existed consistently in the `test` folder. Most modules in this repo, especially in the test folder are not really helpful with their naming, and it felt natural to name a module after the only function it exports.
But I don't have any concerns with renaming it.

karfau:
I temporarily block edits for this PR since I have quite some changes locally already.

> I am now wondering if there could be any way to synchronously process contents from a zip file?

Yes, of course, I know at least one "ugly" way: extract to disk (e.g. in `postinstall` script and use `fs.readSync` in the tests.

But I wanted to 
1. avoid extracting it, to avoid "breaking the license" by "distributing the zip file in a different way"
2. not create redundant state.

So I tried to read from the zip file as art of the tests, which I only found an async way to do. I will research if there is a sync lib for doing that.

brodybits:
> I temporarily block edits for this PR since I have quite some changes locally already.

Makes sense. I am assuming (and hoping) you will also rebase (or merge) to resolve the conflicts:)

> "ugly" way: extract to disk (e.g. in `postinstall` script and use `fs.readSync` in the tests.

I would be fine if we do this into some kind of a temp directory such as `.tmp` or `.temp` and add it to `.gitignore`. I think this should be able to unblock progress without any issue with the license.

That said, it would definitely be nice if we can find a better solution.

I would love it if we could get this fixed in Jest somehow. Probably a long shot.

brodybits:
Hey how about this one: https://nodejs.org/api/zlib.html#zlib_zlib_unzipsync_buffer_options

karfau:
I had a nice idea over night and it works quite well. SO I don't need to look into other options.
I think this PR will be ready in some days.

test/createParser.js
karfau: Thinking about it a bit longer: maybe I just didn't get your point about "overriding".
If my reply was not satisfying, please let me know and elaborate your concerns.

The renaming took place, please check if this is what you had in mind.

karfau:
> I am also wondering if we should raise bug reports for the errors we encounter between xmldom and xmltest?

I'm very satisfied with the current status, so I can go ahead and file that issue, and can refer to it in the comment.

babel.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+	presets: [
+		[
+			'@babel/preset-env',
```
brodybits, line 4: Would you mind adding an explanation why this configuration is needed?

jest.config.js
```diff
@@ -116,7 +116,7 @@ module.exports = {
 
   // A list of paths to directories that Jest should use to search for files in
   // roots: [
-  //   "<rootDir>"
+  //   "test"
```
brodybits, line 119: I think this is not needed, see next comment.

jest.config.js
```diff
@@ -156,8 +156,8 @@ module.exports = {
 
   // The regexp pattern or array of patterns that Jest uses to detect test files
   testRegex: [
-  	/\.test\.js?$/
-	],
+    /test\/.*\.test\.js$/
+  ],
```
brodybits, line 160: I think these changes are not really needed. I was able to run Jest with these changes removed.

stryker.conf.json
```diff
@@ -16,6 +16,9 @@
   },
   "coverageAnalysis": "all",
   "commandRunner": {
-    "command": "npm run test:unit"
-  }
+    "command": "jest"
+  },
+  "files": [
+    "test/**/*.test.js", "test/**/*.xml", "test/**/*.html", "test/createParser.js", "lib/*.js", "package.json", "jest.config.js"
+  ]
```
brodybits, line 23: I think these changes are not needed. We should run Stryker with the entire test suite.

stryker.jest.conf.json
```diff
@@ -0,0 +1,31 @@
+{
```
brodybits, line 1: I think this whole file is not really needed.

test/configure-test-parser.js
```diff
@@ -0,0 +1,38 @@
...
+ * 			  		 default is an empty object
+ *
+ * @param options {{
+ * 					errorHandler?: function (key: ErrorLevel, msg: string),
+ * 					errors?: Partial<Record<ErrorLevel, string[]>>,
+ * 					locator?: Object
+ *				}}
+ * @returns {{parser: DOMParser, errors: Partial<Record<ErrorLevel, string[]>>}}
+ */
+function configureTestParser ({errorHandler, errors = {}, locator = {}} = {}) {
```
brodybits, line 25: As a nit, I would rename this to something like getTestParser.

test/xmltest/not-wf.test.js
```diff
@@ -0,0 +1,31 @@
...
+		xmltest.FILTERS.xml
+	)
+  Object.entries(entries).forEach(([pathInZip, filename]) => {
+
+    if (filename === '145.xml' || filename === '146.xml') {
+      test.todo(`should match ${filename}, has strange chars that fail in snapshot`)
+      return;
+    }
+
+    test(`should match ${filename} with snapshot`, async () => {
```
brodybits, line 17: nit: alignment does not look right in this test

test/xmltest/valid.test.js
```diff
@@ -0,0 +1,27 @@
...
+describe('xmltest/valid', () => {
+  describe('standalone', () => {
+		const entries = xmltest.getEntries(
+			xmltest.FILTERS.VALID.SA.files,
+			xmltest.FILTERS.xml
+		)
+    Object.entries(entries).forEach(([pathInZip, filename]) => {
+
+      test(`should match ${filename} with snapshot`, async () => {
+				const input = (await xmltest.getContent(pathInZip))
```
brodybits, line 14: nit: alignment does not look right in this test

brodybits (commented):
I left some questions and comments. In general:

- `test/xmltest/__snapshots__/not-wf.test.js.snap` seems to be a binary file
- When I did `npm install` on my mac, it seems to remove some entries from `package-lock.json`. I will probably regenerate `package-lock.json` anyway before release, but was wondering about this.

karfau:
Sorry for wasting some of your time since you looked at it so fast. I found out this morning that it was not totally ready.

But I can already reply to some of your comments:

> test/xmltest/__snapshots__/not-wf.test.js.snap seems to be a binary file

No,  It's a [snapshot file from jest](https://jestjs.io/docs/en/snapshot-testing). I'll tell git(hub) to display their diff.

> package-lock.json

I'll check.

stryker.conf.json
karfau: This was here for two reasons:
- improve performance of copying into `.stryker-tmp` folders (I don't think it's needed anymore)
- My expectation is that the xmltest suite covers everything (and more) that the vows testsuite does. I'll double check. If the current score of 99.7% is increased by also running the vows tests as part of stryker, I would of course include them. 

That being said, using npm to run jest and vows will be a major performance drain for running stryker.

stryker.jest.conf.json
karfau: I was trying to run stryker using the "native" `jestRunner` for improved performance. But I was not able to make it work, so I agree it should not be part of this PR.

jest.config.js
karfau: sometimes stryker leaves a `.stryker-tmp/...` directory after running it (reproducible when hitting `Ctrl+C`), running jest after that will also run the tests in those directories.

Also it just speeds up jest since it doesn't need to search for tests anywhere else.

babel.config.js
karfau: I still need to try if it works without this, in which case I'll of cause remove it.

karfau: Not needed and is removed.

stryker.conf.json
karfau: The result is outside of how I understood stryker so far:

stryker command | Mutation score | # Killed | # Survived | # Timeout | # No coverage | # Ignored | # Runtime errors | # Compile errors | Total detected | Total undetected | Total mutants
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
`jest` | 99.70% | 1958 | 6 | 35 (timeoutMS:15s) | 0 | 0 | 0 | 0 | 1993 | 6 | 1999
`npm run test:unit` | 64.73% | 1243 | 705 | 51 (timeoutMS:20s) | 0 | 0 | 0 | 0 | 1294 | 705 | 1999

The higher amounts of timeout is most likely related to tests taking to long (I have see this number go up drastically when the timeout it to low.
But why would more tests result in less coverage?

I will try to see if that is an issue with `npm-run-all` ...

jest.config.js
karfau: sometimes stryker leaves a `.stryker-tmp/...` directory after running it (reproducible when hitting `Ctrl+C`), running jest after that will also run the tests in those directories.

Also it just speeds up jest since it doesn't need to search for tests anywhere else.

brodybits:
Thanks. I may need a few days to finish reviewing.

Unfortunately git and GitHub seem to continue to treat `test/xmltest/__snapshots__/not-wf.test.js.snap` like a binary file. I think `.gitattributes` is not right. I tried fixing it like this, but it still treats the snapshot like a binary file:

```
# jest snapshots can and should show in diff, since they are not binary
*.test.js.snap text
```

When I tried running `test/xmltest/not-wf.test.js` with this change:

```diff
--- a/test/xmltest/not-wf.test.js
+++ b/test/xmltest/not-wf.test.js
@@ -24,7 +24,7 @@ describe('xmltest/not-wellformed', () => {
                                // for 050.xml the result is undefined so `.toString()` needs to be careful
                                const actual = parser.parseFromString(input)
 
-                               expect({actual: actual ? actual.toString() : actual, errors}).toMatchSnapshot()
+                               expect({input, actual: actual ? actual.toString() : actual, errors}).toMatchSnapshot()
                        })
                })
        })
```

then I could see from the snapshot some inputs that result in some non-text outputs:

- `&#RE;`
- `&#X58;`
- `&#0;`
- `&#31;`
- `&#xFFFF;`

and an input that results in an error:

- `&e;`

I am thinking that we should replace the few troublesome non-text output sequences and be able to get rid of `.gitattributes`.

test/xmltest/not-wf.test.js
```diff
@@ -0,0 +1,31 @@
...
+				const input = (await xmltest.getContent(pathInZip))
+					// TODO: The DOCTYPE totally confuses xmldom :sic:
+					// for now we remove it and any newlines after it so we have reasonable tests
+					.replace(/^<!DOCTYPE doc \[[^\]]+]>[\r\n]*/m, '')
+				const {errors, parser} = getTestParser()
+
+				// for 050.xml the result is undefined so `.toString()` needs to be careful
+				const actual = parser.parseFromString(input)
+
+				expect({actual: actual ? actual.toString() : actual, errors}).toMatchSnapshot()
```
brodybits, line 27: I would do something like this to give us cleaner output (not tested):

```suggestion
				const output = actual ? actual.toString() : actual

				if (errors.keys.length() === 0)
					expect({output}).toMatchSnapshot()
				else
					expect({errors, output}).toMatchSnapshot()
```

or maybe this to show the input as well (also not tested):

```suggestion
				const output = actual ? actual.toString() : actual

				if (errors.keys.length() === 0)
					expect({input, output}).toMatchSnapshot()
				else
					expect({errors, input, output}).toMatchSnapshot()
```

test/xmltest/valid.test.js
```diff
@@ -0,0 +1,26 @@
...
+					// TODO: The DOCTYPE totally confuses xmldom :sic:
+					// for now we remove it and any newlines after it so we have reasonable tests
+					.replace(/^<!DOCTYPE doc \[[^\]]+]>[\r\n]*/m, '')
+				const expected = await xmltest.getContent(xmltest.RELATED.out(pathInZip))
+				const {errors, parser} = getTestParser()
+
+				const actual = parser.parseFromString(input).toString()
+
+				expect({actual, errors, expected}).toMatchSnapshot()
```
brodybits, line 22: I would apply similar changes to what I proposed for `test/xmltest/not-wf.test.js` to give us some cleaner output.

I am not 100% sure whether or not we should show the input in the snapshots. It would be more clutter but could also better show what is going on per case.

brodybits:
If we would remove `test/xmltest/__snapshots__/not-wf.test.js.snap` and update `test/xmltest/not-wf.test.js` to skip for a few more file names, then git will treat the snapshot like a text file.

```diff
--- a/test/xmltest/not-wf.test.js
+++ b/test/xmltest/not-wf.test.js
@@ -9,7 +9,10 @@ describe('xmltest/not-wellformed', () => {
 		)
 		Object.entries(entries).forEach(([pathInZip, filename]) => {
 
-			if (filename === '145.xml' || filename === '146.xml') {
+			if (filename === '009.xml' || filename === '093.xml' ||
+			    filename === '142.xml' || filename === '143.xml' ||
+			    filename === '144.xml' ||
+			    filename === '145.xml' || filename === '146.xml') {
 				test.todo(`should match ${filename}, has strange chars that fail in snapshot`)
 				return
 			}
```

(It would be awesome if you could keep these file names in a list.)

I would favor that we proceed with this kind of solution now and track the TODO in an issue.

karfau:
> then I could see from the snapshot some inputs that result in some non-text outputs:
> 
> `&#RE;`
> `&#X58;`

Although the other signs (3-5) are quite obvious, the first two are unclear for me: 
The second one is an `X` and the I have no idea what the first one is referring to.

It was quite a busy week, but I'm able to continue working on it now.

karfau:
I applied all changes that I could derive from your comments.
The only remaining question is the one about still running vows tests when doing stryker.

brodybits:
Hi would you mind if we move the changes to `jest.config.js` and `stryker.conf.json` out for separate PR?

From a quick look I think everything else should be good. I hope to do the final check this weekend and merge soon.

brodybits:
I am also thinking to switch from `stryker.conf.json` to `stryker.conf.js`, where we would be able to add some comments about what is going on with the coverage.

Thanks again for all of your efforts.

karfau:
> Hi would you mind if we move the changes to `jest.config.js` and `stryker.conf.json` out for separate PR?

"Done",at least it's reverted in this branch. (The changes are still present in https://github.com/karfau/xmldom/tree/xmltest-jest-stryker for later reference.)

brodybits (commented):
I pushed a few more updates to remove `expected` from the snapshots, further minimize the snapshots, additional minor cleanup, and nits.

I would rather not keep `expected` in the snapshots since it comes directly from the xmltest (open for discussion).

Also, I did add some semicolons to increase consistency with the rest of the code base. When we do stricter lint, I think it will be less drastic to add "missing" semicolons than remove existing semicolons. (Also open for discussion.)

Please feel free to remove any updates that you think should not be included.

I did leave comments for 2 more things that I think need to be listed in one or more followup issues.

test/xmltest/__snapshots__/not-wf.test.js.snap
```diff
@@ -0,0 +1,1407 @@
...
+
+exports[`xmltest/not-wellformed standalone should match 167.xml with snapshot 1`] = `
+Object {
+  "actual": "<doc>{!fffe!}</doc>",
+}
+`;
+
+exports[`xmltest/not-wellformed standalone should match 168.xml with snapshot 1`] = `
+Object {
+  "actual": "<doc>���</doc>",
```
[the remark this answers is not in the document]
karfau: I checked those: those are chars that are supposed to look like this. So I don't think there needs to be a follow up.

test/xmltest/__snapshots__/not-wf.test.js.snap
```diff
@@ -0,0 +1,1407 @@
...
+
+exports[`xmltest/not-wellformed standalone should match 008.xml with snapshot 1`] = `
+Object {
+  "actual": "<doc>&amp;.entity;</doc>",
+}
+`;
+
+exports[`xmltest/not-wellformed standalone should match 009.xml with snapshot 1`] = `
+Object {
+  "actual": "<doc>{!0!}</doc>",
```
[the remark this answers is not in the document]
karfau: The issue is that putting them like this might again render them as "invisible" depending on what engine displays them. I explicitly rendered them to "ask for attention".

karfau:
The following was my thinking about the stuff that I left in the snapshots:
- `actual` so it's persisted what the actual was when the test was green -> if something breaks, you will see the diff in the output of the failed tests and the git diff in the snapshot file as part of the PR. (So you could say it is there to optimized for a minimal diff in case of updates.)
- `expected` since we are doing replacement of special chars, it is in deed interesting to see how far away xmldom is from what the xmltest cases define as "canonical form". And it was only included when it's different from the actual, so no redundant strings. Regarding "directly coming from the zip file": I good an OK from the author for publishing the xmltest lib to npm and would argue that storing a version with replacements that it failed to compare against, is part of "for internal testing". Also one main argument in my opinion is that there are very few cases that are an exact match, and it's a fast way to "document" the difference.

But I actually discovered that you fixed what was compared: in my last version the expected without replacement was compared to the actual with replacements, which is of course wrong.

So I would strongly argue for adding back `actual`. If you are still concerned about expected, I would still add something that can easily be enabled locally, to more easily see what is going on.

karfau:
> Also, I did add some semicolons to increase consistency with the rest of the code base. When we do stricter lint, I think it will be less drastic to add "missing" semicolons than remove existing semicolons. (Also open for discussion.)

I don't mind about those, of course I would prefer to do it right right away, so ideally there is some eslint/prettier config that already suggests to put them.

brodybits:
> * By removing expected from the snapshot, a test no longer fails if actual is currnelty the same as expected and later becomes different from expected!

Unfortunately I do not really understand that statement. If actual becomes different from expected, wouldn't it cause the snapshot to change?

karfau:
> > * By removing expected from the snapshot, a test no longer fails if actual is currnelty the same as expected and later becomes different from expected!
> 
> Unfortunately I do not really understand that statement. If actual becomes different from expected, wouldn't it cause the snapshot to change?

Hm, I think it only makes sense when actual is already part of the snap. That's most likely why I came to that conclusion.

brodybits:
I think I am ready to merge this one, if my latest changes are OK with @karfau. I pushed some changes to be more consistent with the Prettier formatting that I recently proposed in PR #120.

I am still not 100% happy with keeping the expected output from `xmltest`, since this is coming from the ZIPped test suite. But the rationale of tracking expected vs actual output makes sense to me.

I did also notice JSDoc typecheck info in `test/get-test-parser.js`, guessing it is there to make VSCode happy. I would like to use JSDoc comments for type checking of our JavaScript with help from `tsc` at some point in the future, similar to what I have contributed to Prettier.

karfau:
I'm very happy about your "pretty" changes. :+1: for applying prettier.

In my IDE of choice (IntelliJ IDEA) I get more reasonable warnings when adding the types and I have also used it with tsc before. Makes totally sense to me.
Also applying that to the actualt API of the lib (not just in tests) would make it accessible for people that use the lib and any IDE that considers them. So it's half way for great (and checked) docs.
So at least for me it's not "about making VSCode happy", but if it helps with that, I'm even more happy :)

Feel free to land it :tada:

karfau:
BTW: the xmltest suite is like a "shotgun" (where it's bullets always hit the same spots). 
It was a fast approach to make sure stuff doesn't break in the future without us noticing.

Once we have more fine grained tests we might consider dropping it again (or relying less on it).

karfau:
Just remembered to sneak in the latest version of xmltest, and it also comes from npm now. So any future updates will be taken care of by renovate.

brodybits:
I added one more blank line, which I had forgot to push before, and merged with a hopefully properly cleaned up commit message.

Yes it definitely makes sense to start with the fast approach and transition to finer-grained testing over time.

I am also thinking it may not be so bad to just keep the `xmltest` cases as a form of integration testing.

karfau:
:tada: 
Thx a lot for the collaboration.

Issue #71, opened 2020-06-29
Allow custom entityMap to be passed as DOMParser option


brodybits:
The changes related to bug #57 (broken `&nbsp;` functionality), and lack of coverage of `lib/entities.js` detected by Stryker (see PR #70), got me thinking if we could import and reuse a better-tested implementation from npm. For example, with help from Google "npm html entities" & "npm xml entities":

- https://www.npmjs.com/package/html-entities
- https://www.npmjs.com/package/entities

and one step further to import SAX parser, as I suggested in #55?

**Addition by karfau**:

I think that we are currently quite clear that we do not want to introduce a dependency.

I think it would be worth while to allow users to override the entityMap (or a method that takes care of it) as an option for the DOMParser.
Currently xmldom applies different entity mapping for XML and HTML parsing which I think is not documented anywhere.

But of course we will also have to think about how it can be applied to serialization...

I'm aware that at least one downstream is reimplementing the replacement to apply it's mechanism before parsing with xmldom and after serialization, which is of course quite a tricky task and requires to wrap the XMLDOM API or play with the prototypes. We should just keep this in mind when implementing anything, so that similar approaches can reduce the amount of code they have to maintain.

karfau:
I don't think it's a good reason to introduce a dependency.
Since we are currently trying to support ES5 without any build chain, we would need to find something that also supports that and make sure it doesn't stop doing that.
But it might still be worth investigating.
(I would recommend https://npms.io/search?q=html+entities and https://npms.io/search?q=entities as a starting point for research.)

karfau:
I think that we are currently quite clear that we do not want to introduce a dependency.

I think it would be worth while to allow users to override the entityMap (or a method that takes care of it) as an option for the DOMParser. 
Currently xmldom applies different entity mapping for XML and HTML parsing which I think is not documented anywhere.

But of course we will also have to think about how it can be applied to serialization...

I'm aware that at least one downstream is reimplementing the replacement to apply it's mechanism before parsing with xmldom and after serialization, which is of course quite a tricky task and requires to wrap the XMLDOM API or play with the prototypes. We should just keep this in mind when implementing anything, so that similar approaches can reduce the amount of code they have to maintain.

brodybits:
Makes sense to me, thanks.

+1 for user-pluggable method for entity mapping, thinking further if we could have a user-pluggable SAX parser?

karfau:
[...]

PR #113, opened 2020-08-25
fix: typecheck source param before parsing

### LINKED ISSUE #98: Infinite-loop error
When passing an object to DOMParser's `parseFromString` you'll get infinetely looping error:
```js
const {DOMParser} = require('xmldom');
new DOMParser().parseFromString({});
```
Obviously you shouldn't pass an object there, but still this should be fixed.

jesse-y:
Updates `DOMParser.parseFromString` to also check that the `source` provided is a string. To prevent sax from throwing undefined errors when calling `indexOf` on the source.

Adds unit tests to `test/parse/simple.vows.js` to cover this edge case.

Resolves #98

### DIFF
--- a/lib/dom-parser.js
+++ b/lib/dom-parser.js
@@ -21,7 +21,7 @@ DOMParser.prototype.parseFromString = function(source,mimeType){
 		defaultNSMap['']= 'http://www.w3.org/1999/xhtml';
 	}
 	defaultNSMap.xml = defaultNSMap.xml || 'http://www.w3.org/XML/1998/namespace';
-	if(source){
+	if(source && typeof source === 'string'){
 		sax.parse(source,defaultNSMap,entityMap);
 	}else{
 		sax.errorHandler.error("invalid doc source");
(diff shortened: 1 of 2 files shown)


brodybits:
Resolution looks OK to me, thanks.

We may rework the test cases in the future, as discussed in #71. But I think we may as well merge now to resolve the issue. Any comments from @karfau?

jesse-y:
Yeah, I'd be fine with these tests being rewritten in another framework (I'm more familiar with Jest, so it's a +1 for me in regard to #72 ). Also happy to help on that front when some of the other housekeeping PRs (#102 #110 ) are in.

karfau (approved):
Sorry for the delay, go for it!

karfau:
FYI: 0.9.0 will change this to be a `fatalError`, which causes `DOMParser.parseFromString` to throw by default. This change helps to avoid the situation, in which `undefined` is returned instead of a `Document`.

PR #114, opened 2020-08-26
test: Add jest and first tests

karfau:
as discussed in #72 and #112 so we can get started using jest for new tests.

- `npm run test` run everything there is
- `npm run jest` just runs the jest tests (`test/*.test.js`)

(The tests have been picked from cases that stryker identified in my run in #112 .)

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -21,7 +21,7 @@ module.exports = {
   // collectCoverageFrom: undefined,
 
   // The directory where Jest should output its coverage files
-  coverageDirectory: "coverage",
+  // coverageDirectory: "coverage",
 
   // An array of regexp pattern strings used to skip coverage collection
   // coveragePathIgnorePatterns: [
@@ -29,7 +29,7 @@ module.exports = {
   // ],
 
   // Indicates which provider should be used to instrument code for coverage
-  coverageProvider: "v8",
+  // coverageProvider: "v8",
 
   // A list of reporter names that Jest uses when writing coverage reports
   // coverageReporters: [
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,0 +1,194 @@
+// For a detailed explanation regarding each configuration property, visit:
+// https://jestjs.io/docs/en/configuration.html
+
+module.exports = {
+  // All imported modules in your tests should be mocked automatically
+  // automock: false,
+
+  // Stop running tests after `n` failures
+  // bail: 0,
+
+  // The directory where Jest should store its cached dependency information
+  // cacheDirectory: "/tmp/jest_rs",
+
+  // Automatically clear mock calls and instances between every test
+  clearMocks: true,
+
+  // Indicates whether the coverage information should be collected while executing the test
+  // collectCoverage: false,
+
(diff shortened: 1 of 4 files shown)


jest.config.js
```diff
@@ -0,0 +1,194 @@
...
+  // The directory where Jest should output its coverage files
+  coverageDirectory: "coverage",
+
+  // An array of regexp pattern strings used to skip coverage collection
+  // coveragePathIgnorePatterns: [
+  //   "/node_modules/"
+  // ],
+
+  // Indicates which provider should be used to instrument code for coverage
+  coverageProvider: "v8",
```
brodybits, line 32: I think the `coverageDirectory ` and `coverageProvider` options are not needed, since we are using Stryker rather than another tool to measure our code coverage.

package.json
```diff
@@ -29,8 +29,9 @@
     "test": "npm-run-all test:assert:strict test:unit lint",
     "test:assert": "vows test/assert.vows.checks.js",
     "test:assert:strict": "XMLDOM_ASSERT_STRICT=1 npm run test:assert",
-    "test:unit": "npm run test:vows",
-    "test:vows": "vows 'test/**/*.vows.js'"
+    "test:unit": "npm-run-all test:vows jest",
+    "test:vows": "vows 'test/**/*.vows.js'",
+    "jest": "jest"
```
brodybits, line 34:
```suggestion
    "test:unit": "npm-run-all test:vows test:jest",
    "test:vows": "vows 'test/**/*.vows.js'",
    "test:jest": "jest"
```

I would normally rather keep it sorted alphabetically, but maybe it is good to keep more general entries before more specific entries.

test/dom-parser.test.js
```diff
@@ -0,0 +1,80 @@
...
+			const XML = '<xml attr="&quot;">&lt; &amp;</xml>'
+			expect(
+				new DOMParser().parseFromString(XML).toString()
+			).toEqual(XML)
+		})
+
+		test('should provide access to textContent and attribute values', () => {
+			// provides an executable example for https://github.com/xmldom/xmldom/issues/93
+			const XML = `
+			<pdf2xml producer="poppler" version="0.26.5">
```
brodybits, line 51: As a nit, I think this will give us a string with an extra blank line. I think it would be ideal to use something like dedent or outdent. But probably not worth changing in this PR.

karfau:
I totally agree on your comments. Maybe I should rather have put the "Todo/Fixme comments into this PR instead of in the code, since my intention was not to land them, but to get feedback on them.
Will Do it differently next time.

brodybits:
> my intention was not to land them, but to get feedback on them

Got it now, that was not clear to me when I reviewed the changes.

I did want to get this PR finished and merged pretty quickly in order to unblock other progress, considering that I think test coverage is way more important than dealing with some other loose ends. That is what I tried to say by my nits "not worth fixing in this PR".

In case we do expect any loose ends in a PR, I think it would be a nice bonus to raise an issue. But comments like TBD, TODO, FIXME are fine with me as well if they make sense.

Thanks again for your contributions.

PR #116, opened 2020-09-02
Add `entries` method to `NodeList`

colejhudson:
Hi there! This PR adds support for an [`entries`](https://developer.mozilla.org/en-US/docs/Web/API/NodeList/entries) method ([here](https://github.com/colejhudson/xmldom/blob/add-entries-method-to-nodelist/lib/dom.js#L109-L118)) on `NodeList`, as mentioned in the 'methods' section on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/NodeList). More generally, it makes `NodeList` objects iterable by adding a `Symbol.iterator` property.

So far as I could tell, it didn't seem like there was a central place for testing DOM objects, so I created a [`dom.test.js`](https://github.com/colejhudson/xmldom/blob/add-entries-method-to-nodelist/test/dom.test.js) file in the `test` directory to host the tests for this change.

As an aside, this feature may be beyond the scope of this project if you're trying to stick to the commonly accepted specs. Despite it's inclusion in the MDN, I couldn't find any mention of an `entries` method in either of [WHATWG](https://dom.spec.whatwg.org/#interface-nodelist)'s or [W3C](https://www.w3.org/TR/DOM-Level-3-Core/core.html#ID-536297177)'s specs.

Thanks!

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -4,7 +4,7 @@ env:
   es2020: true
   node: true
 plugins:
-  - es5
+  - es6
 # FUTURE TBD:
 # extends: 'eslint:recommended'
 parserOptions:
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -4,7 +4,7 @@ env:
   es2020: true
   node: true
 plugins:
-  - es6
+  - es5
 # FUTURE TBD:
 # extends: 'eslint:recommended'
 parserOptions:
--- a/lib/.eslintrc.yml
+++ b/lib/.eslintrc.yml
@@ -1,2 +1,0 @@
-extends:
-  - 'plugin:es5/no-es2015'
--- a/lib/.eslintrc.yml
+++ b/lib/.eslintrc.yml
@@ -1,0 +1,2 @@
+extends:
+  - 'plugin:es5/no-es2015'
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -106,6 +106,16 @@ NodeList.prototype = {
 	item: function(index) {
 		return this[index] || null;
 	},
+	[Symbol.iterator]: function* () {
+		for (let i = 0; i < this.length; i += 1) {
+			yield this.item(i);
+		}
+
+		return this.length;
+	},
+	entries: function() {
+		return this[Symbol.iterator];
+	},
 	toString:function(isHTML,nodeFilter){
 		for(var buf = [], i = 0;i<this.length;i++){
 			serializeToString(this[i],buf,isHTML,nodeFilter);
--- a/package-lock.json
+++ b/package-lock.json
@@ -2472,14 +2472,11 @@
         }
       }
     },
-    "eslint-plugin-es6": {
-      "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/eslint-plugin-es6/-/eslint-plugin-es6-1.0.0.tgz",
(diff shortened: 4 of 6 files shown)


brodybits (changes requested):
Thanks for the contribution. The proposed switch to ES6 would break usability in some older browsers, see PR #107.

I also do not see where the `entries` member function itself is tested.

colejhudson:
Hey Brody, thanks for the review.

I didn't explicitly add a test for the `entries` method because it acts as a wrapper around the `Symbol.iterator` property. So my tests target that property specifically in an attempt to be more general. If it's prefered, I can totally add an explicit test for `entries` later tonight.

I'll see if I can't get this working in es5 then too.

brodybits:
From this document on NodeList: https://developer.mozilla.org/en-US/docs/Web/API/NodeList

I think we should add the `entries` and `forEach` methods.

With ES2015 we could implement `entries` like this:

```js
  entries: function() {
    return Array.from(this).entries();
  },
```

But to be extra-safe for older ES5 browsers, I think we should implement `entries` like this (not tested):

```js
  entries: function() {
    var a = [];
    for (var i=0; i<this.length; ++i) a.push(this[i]);
    return a.entries();
  },
```

It would be awesome if you could add the explicit test of the `entries` method.

I am not convinced we should add the method for `Symbol.iterator`, since I do not see it in the MDN document, but we could probably implement it in a manner like this, as a wrapper around the new `entries` method:

```js
NodeList.prototype[Symbol.iterator] = function() {
  return this.entries();
};
```

I would be equally happy to include adding `forEach` in this PR or leave it for future consideration.

karfau:
@colejhudson Thank you for investing the time into this contribution already. Do you still plan to work on this and apply the changes that have been requested?

I don't see us (maintainers) investing time into applying changes to this code in the current state soon, since I'm currently not aware of any spec that defines them.

Although I think the topic is worth discussing or keeping in mind, from the perspective of maintainers investing time into it, I rather see it as something for after xmldom reaches 1.0.

So if you prefer to close this PR, maybe it makes sense to rather create an issue or a discussion for it?

(If there is no reaction to this within the next month or two I will most likely close this PR.)

colejhudson:
Hey @karfau, I appreciate the ping and apologize for letting this get so stale; I had forgotten all about it! I'll go ahead and close this PR as I won't have the time to work on it.

I appreciate all the hard work you guys do, cheers!

wh1t3cAt1k:
I would like to submit a provocative proposal and to reopen & merge this PR as is, dropping ES5 support.

image

Microsoft has officially ended support for IE11 almost a year ago, and it was one of the last guys to only partially support ES6.

karfau:
This is not just about ES5 or browsers, so yes it's a provocative suggestion.
Let's continue the the discussion regarding the runtimes we should support in #214 

And since I didn't pull that branch of @colejhudson to my disk back then, I'm not able to recover the diff to judge if it would be possible. If somebody has it, feel free to create another PR and link to this one.

A lot of things have happened since then, so I assume there would most likely be some conflicts.

PS: we don't even have a major version yet, so it would be a minor bump. But a massive one. If you are up for supporting such a change (which most likely also involves switching the code in this repo from commonjs to something else?) let me know and we can have a conversation about the cost of change, the path towards this goal, possible contributions along the way, etc.
And just to make this clear: I'm not saying it won't happen, but this is currently a one-man-show and I"m not going to take care of that without (a decent amount of) support.

PR #120, opened 2020-09-15
apply Prettier formatting with no semicolons

brodybits:
Given that Prettier formatting would touch most lines of the code anyway, as we can see in PR #110, my idea is now to remove semicolons at the same time.

I would continue to favor improving test coverage, as proposed in PR #112, before merging this one.


brodybits:

PR #121, opened 2020-09-16
Port Vows tests to Jest & remove test:vows script

brodybits:
I have done a very simple port of the Vows tests to Jest.

The one exception is `test/assert.vows.checks.js`, which I think can be removed by converting the custom assertions to more standard, strict assertions like @awwright started in PR #89.

I think this could be another positive step towards the more future-proof testing strategy discussed in #72 and avoid issues like <https://github.com/stryker-mutator/stryker/issues/2459> with Stryker.


karfau (approved):
This looks awesome!

I agree as a next step we should convert the custom assertions to jest's `expect` and  get rid if the whole complicated assertion stuff. But rather in a separate PR, otherwise git might no longer be able to treat those files as moved.

karfau:
One silly question: did you run stryker with this branch? What is the result?

brodybits:
Thanks for the quick review.

No I did not try with Stryker. My machine is way too slow to handle it and I was having some issues with cloud access.

I am watching the Stryker run on GitHub now. Good news is that the score seems to be above 80% now. The bad news is that it seems to be quite a bit slower. We should probably switch over to Jest runner.

I am now very hopeful that the efforts here and in #112 can help unblock further cleanup and upcoming contributions.

brodybits:
FYI the Stryker score went from 64.7% to 64.8% with this change.

I did miss one thing that I fixed in PR #122. Thanks to @karfau for the quick review on that as well.

karfau:
I think the migration to jest is still incomplete, but I'm not sure what the next step should be:

- There are tons of console output added by the legacy tests, it really makes it hard to understand what is actually going wrong.
- One source of log output is the warning about strict check would be failing from our custom assertion. All cases I found so far just need an added `+''` (or `.toString()`) to remove those warnings.
- The tests for the custom assertions are the only ones running on vows by now:
  - should we drop them completely and convert all tests to `expect` assertions right away?
  - or would you prefer to keep them and also move it's tests to jest?

I'm happy to take over some part of it to get rid of vows, but we should be clear about order and who is responsible for which part, to avoid "stepping on each others toes".
(We could also continue this discussion in #72.)

brodybits:
Yup. My primary purpose was to help unblock some needed progress. I would definitely favor that we continue the discussion in #72.

In terms of priorities, I think people should just work on whatever is most important to them in general. To avoid stepping on each other's toes, I would favor that whoever starts working on something raise a draft PR.

Thanks again for all of your contributions, much appreciated !!

Issue #99, opened 2020-08-12
0.4.0 release in 3Q 2020


brodybits:
__updated:__

I think we should make the new 0.4.0 release soon and adjust the planning of the following milestones:

- https://github.com/xmldom/xmldom/milestone/1 (0.4.0) - push remaining items back to 0.5.0
- https://github.com/xmldom/xmldom/milestone/3 (0.5.0) - consider moving these into a 0.6.0 release

__Added item: update changelog (issue #62)__

~~I would like to get a couple more changes in before making the release, if they don't take too long:~~

- ~~PR #36 (cleaner module exports in index.js)~~
- ~~fix main package entry (issue #53)~~

/cc @karfau @kethinov @timbru31 please speak up asap in case of any concerns

karfau:
DO I understand correctly, that the only changes you want to have as part of 0.4.0 are the ones listed in this description and the other open ticket would be moved to "a new 0.5.0"?

I am working on a much better test coverage on my machine using jest, there shouldn't be much stuff to do to have a PR ready. It's up to you to say whether this should be "part of the release" or it's not related.
I'm also able to invest more time again from this week on (my relocation and vacation is "complete").

brodybits:
That would be really awesome if you have the improved test coverage in progress. I would also not mind starting with an incomplete, draft PR if you like.

I am actually pretty busy with some other project in parallel so there is not a really super rush on my part. I was just triggered by both the number of months and the number of valuable fixes that have been made since the last release.

karfau:
I think I can commit to file at least a draft PR latest at the weekend. I will describe the issues I had with this approach in the PR then.

brodybits:
FYI some recent contributions do not seem to be very well tested in the mutation report:

- getElementsByClassName - updated by PR #91
- code affected by PR #100

brodybits:
I think we should try to make this release soon. It would be ideal if we could update the changelog as discussed in issue #62.

brodybits:
@karfau I would like to make the release this week if there is nothing else missing.

karfau:
When you say "this week", do you mean until today or until 4th of October?
In the latter case  I'll try to finish the PR related to the Changelog asap.
I'm not able to commit to any thing relevant to the actual behavior of xmldom within this week.

brodybits:
[...]

PR #123, opened 2020-09-16
apply Prettier formatting --no-semi in test only

brodybits:
This proposal introduces Prettier for formatting with tabs & no semicolons in test only.

I am basically splitting this out from PR #120.

I would like to apply this formatting to the `lib` code as well, once we further improve the test coverage as discussed in #72.

It would be ideal if we could do this before making the 0.4.0 release discussed in #99.


karfau (commented):
I would review this once the conflicts have been resolved. (I think it's important to not change spacing inside inline (multiline) XML test snippets, but I'm not sue such cases exist.)

brodybits:
I just resolved the conflicts (with a little bit of a trick). I am not sure if I follow your concern, would definitely appreciate if you can point out an example in case the unwanted spacing happens. Thanks!

test/3rd-cases/mock.js
```diff
@@ -1,3 +1,3 @@
-'use strict';
+'use strict'
 
-exports.test = 1;
```
karfau, line 3: Nit (unrelated to this PR): I think we should finally get rid if files that are not part of any test suite and where it's not clear what they have ever been good for.
All files in this folder are an absolute riddle to me.

test/assert.js
```diff
@@ -44,28 +47,37 @@ function locateAssertion (error) {
...
+			'(strict ===) ',
+			...messages,
+			caught.message,
+			locateAssertion(caught),
+		]
+		if ((this.strictThrows || assert.strictThrows)()) {
+			caught.message = msg.join(' ')
+			throw caught
+		}
+		;(this.log || assert.log)(...msg)
```
karfau, line 79: I don't know why this semicolon is here. But since we are getting rid of this whole file soon ...

test/dom/element.test.js
```diff
@@ -1,156 +1,183 @@
-'use strict';
+'use strict'
 
-var wows = require('vows');
-var DOMParser = require('../../lib/dom-parser').DOMParser;
+var wows = require('vows')
```
karfau, line 3: Question: I was just surprised to still see those "imports", shouldn't they be gone from the test files by now?

test/error/error.test.js
```diff
@@ -1,10 +1,11 @@
...
-var wows = require('vows');
-var DOMParser = require('../../lib/dom-parser').DOMParser;
+var wows = require('vows')
+var DOMParser = require('../../lib/dom-parser').DOMParser
 const assert = require('../assert')
 
-var xml = '<scxml xmlns="http://www.w3.org/2005/07/scxml" version="1.0"\n\
+var xml =
+	'<scxml xmlns="http://www.w3.org/2005/07/scxml" version="1.0"\n\
        profile="ecmascript" id="scxmlRoot" initial="start">\n\
```
karfau, line 9: This is the case I was talking about before. But it looks like Prettier is smart enough to not touch these spaces.

test/index.js
```diff
@@ -1,67 +1,68 @@
...
-try{
-	var libxml = require('libxmljs');
-}catch(e){
-	var DomJS = require("dom-js");
+var XMLSerializer = require('../lib/dom-parser').XMLSerializer
+var DOMParser = require('../lib/dom-parser').DOMParser
+try {
+	var libxml = require('libxmljs')
+} catch (e) {
+	var DomJS = require('dom-js')
```
karfau, line 8: We should also get rid of this file.

test/parse/locator.test.js
```diff
@@ -1,81 +1,104 @@
...
-].join('\n')
-	var parser = new DOMParser({locator:{}});
-    var doc = parser.parseFromString(xml, 'text/xml');
-   	var trans = doc.getElementsByTagName('transition')[0];
-   	assert.equal(trans.lineNumber , 10)//,''+trans+trans.lineNumber+'/'+trans.parentNode.previousSibling.previousSibling.lineNumber)
-   	
+	it('empty line number', () => {
+		var xml = [
+			'<scxml xmlns="http://www.w3.org/2005/07/scxml" version="1.0"',
+			'       profile="ecmascript" id="scxmlRoot" initial="start">',
```
karfau, line 24: Another (copy?) of the "spaces in string literals" case, but not an issue.

karfau (approved):
I think it's all right.
Looking forward to get rid of all those strange non-sense files.

brodybits:
I think it would be good to raise one or more issues to track the loose ends that we identified here.

I will likely merge this later tonight. Thanks to @karfau for the timely review.

test/assert.js
brodybits: FYI the rationale is given here: https://prettier.io/docs/en/rationale.html#semicolons

test/error/error.test.js
```diff
@@ -16,85 +17,130 @@ var xml = '<scxml xmlns="http://www.w3.org/2005/07/scxml" version="1.0"\n\
...
-    	locator:{},
-    	errorHandler:{
-			error:function(msg){
-				error.push(msg);
-				throw new Error('from throwing errroHandler.error')
+			errorHandler: function (msg) {
+				error.push(msg)
+			},
+		}).parseFromString(faulty, 'text/xml')
+		Array.from(['warn', 'warning', 'error', 'fatalError']).forEach((k) => {
```
brodybits, line 88: FYI I used `Array.from` here, as a trick to avoid the automatic semicolon insertion (ASI) by Prettier.

We should probably add some blank lines to this test case as well.

PR #129, opened 2020-09-18
apply Prettier formatting to lib

brodybits:
- move `extends: plugin:prettier/recommended` to top-level `.eslintrc.yml`
- apply Prettier formatting to `lib`

This is basically a continuation of the Prettier formatting that was already applied to `test` in merged PR #123.

Note that the rationale behind the Prettier options used is given in the updated description of merge PR #123.

- [ ] I would rather have further improvement of test coverage with the testing strategy discussed in #72 before merging this proposal. While it should be safe to let Prettier reformat the code, I think it would be better practice to further improve the Stryker mutation testing score before updating such a large amount of the code.


karfau:
this one has tons of conflicts so I started from scratch in #447 which also includes the new example code.

PR #126, opened 2020-09-17
Completely remove remaining use of Vows

brodybits:
In followup to PR #121 (port to Jest) and related to PR #89 (use mostly strict assertions), this completely removes remaining use of Vows and related package scripts no longer needed.

TODO items:

- [ ] resolve known conflicts since I started working on this before merging PR #124
- [ ] I would like to finish review & merge of PR #89 before merging this one.

/cc @awwright @karfau

PR #138, opened 2020-10-04
Update CHANGELOG.md using auto-changelog

karfau:
I just rebased that branch on master, I think for now it's good enough to tweak the result until we like it.

It's quite some changes.

I already spotted that #126 has a bad title, since it didn't completely remove the remaining use of vows. But maybe this is anyways something to not mention in the changelog?

Feel free to modify the branch or add suggestions to commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - **BREAKING** Typecheck source param before parsing [`#113`](https://github.com/xmldom/xmldom/pull/113)
 - Preserve doctype with sysid [`#144`](https://github.com/xmldom/xmldom/pull/144)
 - Remove ES6 syntax from getElementsByClassName [`#91`](https://github.com/xmldom/xmldom/pull/91)
-- Revert "Add lowercase of åäö in entityMap" [`#84`](https://github.com/xmldom/xmldom/pull/84)
+- Revert "Add lowercase of åäö in entityMap" due to duplicate entries [`#84`](https://github.com/xmldom/xmldom/pull/84)
 - fix: Convert all line separators to LF [`#66`](https://github.com/xmldom/xmldom/pull/66)
 
 ### Docs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,44 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 <!-- auto-changelog-above -->
+
+## [0.4.0](https://github.com/xmldom/xmldom/compare/0.3.0...0.4.0)
+
+### Fixes
+- **BREAKING** Restore `&nbsp;` behavior from v0.1.27 [`#67`](https://github.com/xmldom/xmldom/pull/67)
+- **BREAKING** Typecheck source param before parsing [`#113`](https://github.com/xmldom/xmldom/pull/113)
+- Preserve doctype with sysid [`#144`](https://github.com/xmldom/xmldom/pull/144)
+- Remove ES6 syntax from getElementsByClassName [`#91`](https://github.com/xmldom/xmldom/pull/91)
+- Revert "Add lowercase of åäö in entityMap" [`#84`](https://github.com/xmldom/xmldom/pull/84)
+- fix: Convert all line separators to LF [`#66`](https://github.com/xmldom/xmldom/pull/66)
+
+### Docs
+- Update CHANGELOG.md through version 0.3.0 [`#63`](https://github.com/xmldom/xmldom/pull/63)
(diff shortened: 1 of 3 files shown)


brodybits:
In short, I would try not to include anything related to test, dev dependencies, configuring Renovate, etc.

A few side points:

It would be really awesome if you could check the Renovate updates and merge them assuming they are OK.

Unfortunately I will not have so much time to make the release for the next week or so due to some urgent work. I think this should give us some extra time to see if we can work out using the auto-changelog.

I have also thinking it would be nice to get rid of package-lock.json for multiple reasons:

- more opinionated about using npm vs Yarn
- adds bloat when merging updates such as this one and some other, recent updates updates
- more arguments here: https://www.twilio.com/blog/lockfiles-nodejs

karfau:
@brodybits This PR is just a "heads up" to make it easier to create the release. So please remove anything from the changelog you don't like.

I would still see the maintenance of the changelog as one step of the release process.

And it would also be a valid approach to close this PR and either use a different tool for creating the changelog or hand crafting it.

I just spotted that I mistakenly added `auto-changelog` to the dependencies, to make above points more clear I just removed the dependency from package.json and lock file and invoke it using `npx`.

I'm going to create a new meta issue regarding lock files.

karfau:
> It would be really awesome if you could check the Renovate updates and merge them assuming they are OK.

Landed all beside major stryker upgrade.

karfau:
While the approach that was present before was nice because it was executed as part of bumping the version, it was also hard to keep it up to date. I think it will always need manual tweaks, so I decided to decouple it from the version bump.

Not it's just about calling `nom run changelog` before or as part of releasing and it will add a new "Unreleased" section on top of the Changelog that can be tweaked.

I prepared the one for 0.4.0 and moved it below the "magic comment", running the above command again will add the "Unreleased" section again, so any new noteworthy changes can be moved down.

Hope this improves the situation and we can release 0.4.0 soon. Again: fee free to tweak anything as you like.

CHANGELOG.md
```diff
@@ -1,3 +1,44 @@
...
+<!-- auto-changelog-above -->
+
+## [0.4.0](https://github.com/xmldom/xmldom/compare/0.3.0...0.4.0)
+
+### Fixes
+- **BREAKING** Restore `&nbsp;` behavior from v0.1.27 [`#67`](https://github.com/xmldom/xmldom/pull/67)
+- **BREAKING** Typecheck source param before parsing [`#113`](https://github.com/xmldom/xmldom/pull/113)
+- Preserve doctype with sysid [`#144`](https://github.com/xmldom/xmldom/pull/144)
+- Remove ES6 syntax from getElementsByClassName [`#91`](https://github.com/xmldom/xmldom/pull/91)
+- Revert "Add lowercase of åäö in entityMap" [`#84`](https://github.com/xmldom/xmldom/pull/84)
```
brodybits, line 16: suggested tweak:

```suggestion
- Revert "Add lowercase of åäö in entityMap" due to duplicate entries [`#84`](https://github.com/xmldom/xmldom/pull/84)
```

package.json
```diff
@@ -87,5 +88,12 @@
   "bugs": {
     "url": "https://github.com/xmldom/xmldom/issues"
   },
-  "license": "(LGPL-2.0 OR MIT)"
+  "license": "(LGPL-2.0 OR MIT)",
+  "auto-changelog": {
+    "commitLimit": false,
+    "remote": "upstream",
+    "startingVersion": "0.3.0",
```
brodybits, line 95: Wouldn't we have to update this after every release?

brodybits (commented):
Thanks @karfau. I raised an extra tweak and a question. In general:

I am now thinking to update `CHANGELOG.md` separately, which I will probably do very soon when I make the 0.4.0 release.

I am also part of a probably small minority who do not like using npx to run tools directly from npm. But this is probably better than adding the overhead of another dev dependency.

package.json
karfau: Without adding this `auto-changelog` failed with an error with a not very helpful message, I assume because of all the inconsistencies in earlier releases/tags, etc. Since we don't care how compatible it is to those older releases I just added this to be able to use the tool.

But I just checked again: since the current approach is to just dump the information about the unreleased changes, it' possible to remove this option.

karfau:
> I am also part of a probably small minority who do not like using npx to run tools directly from npm. But this is probably better than adding the overhead of another dev dependency.

I don't have any preference on any of the both options, since I like lock files, but I think it's a valid approach for now. It's basically just documenting how to easily get the list of changes into the changelog and can be replaced with "any other script" later.

I know there are quite some concerns about npx and I also try to use it only if there is a good reason.
And I'm very happy that [npm@v7 integrates the npx binary into it's own codebase where it's just an alias for `npx exec`](https://blog.npmjs.org/post/626173315965468672/npm-v7-series-beta-release-and-semver-major) which will invalidate **some** of the valid concerns out there right now.

karfau:
Recreated the parts that are still relevant in #180

PR #144, opened 2020-10-07
fix: preserve doctype with sysid (#143)

### LINKED ISSUE #143: DOCTYPE SYSTEM identifiers do not appear to be properly parsed
When parsing an XML document that includes a doctype similar to `<!DOCTYPE custom SYSTEM "custom.dtd">` (SYSTEM identifier for a local DTD filename), the system identifier appears to be parsed as ">".  This is causing problems for us in round-trip usage, as parsing and then serializing the document results in corruption of the DOCTYPE.

I plan on creating a pull request with a fix for this issue.

davidmc24:
Fixes #143

### DIFF
--- a/lib/sax.js
+++ b/lib/sax.js
@@ -531,8 +531,16 @@ function parseDCC(source,start,domBuilder,errorHandler){//sure start with '<!'
 		var len = matchs.length;
 		if(len>1 && /!doctype/i.test(matchs[0][0])){
 			var name = matchs[1][0];
-			var pubid = len>3 && /^public$/i.test(matchs[2][0]) && matchs[3][0]
-			var sysid = len>4 && matchs[4][0];
+			var pubid = false;
+			var sysid = false;
+			if(len>3){
+				if(/^public$/i.test(matchs[2][0])){
+					pubid = matchs[3][0];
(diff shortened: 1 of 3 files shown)


brodybits:
@karfau do you think we want to include this in the upcoming 0.4.0 release?

karfau (approved):
Very nice! Thx a lot.

karfau:
@brodybits I don't see a reason why we shouldn't :+1:

brodybits:
The one little thing I wonder is if these changes could introduce any more surviving mutants when running Stryker. Unfortunately I have no time to try this now, and I don't think this should be blocking. @karfau it would be great if you can merge this (as a squash commit). Thanks again!

davidmc24:
I'm glad the PR seems to be well received. Thank you for your work maintaining this library.

karfau:
@brodybits I tried to run stryker after applying the patch, and have received 8 new mutants, but my scores differ for every file from the one produced by github actions on master. So not sure what to derive from it. I guess we will see the result when it lands.

brodybits:
> my scores differ for every file from the one produced by github actions on master

Maybe we should investigate which mutations differ in the reports from the command line vs GitHub.

PR #171, opened 2021-01-13
Fix breaking preprocessors' directives when parsing attributes

yeegor:
### Overview

The issue described above has been found during the development of the [create-scandipwa-app](https://github.com/scandipwa/create-scandipwa-app).

### Motivation

See the following code example. It is located in a PHP file and is valid: the corresponding function is called when the file is parsed. This has been tested using the Magento 2 framework, nonetheless, it should be possible to recreate this without the framework.

```html
<html lang="<?= $this->getLocaleCode() ?>">
```

The code fragment above, when parsed back and forth, is expected to remain the same. But in fact, it turns into the following code.

```html
<html lang="&lt;?= $this->getLocaleCode() ?>">
```

### Context

The used version is `0.4.0`
The parsing process has been invoked as follows:

```js
const xmldom = require('xmldom');

const domParser = new xmldom.DOMParser();
const xmlSerializer = new xmldom.XMLSerializer();

const templateFile = ...

const parseToDOM = (code) => domParser.parseFromString(code, 'text/html');
const parseToString = (dom) => xmlSerializer.serializeToString(dom);

// This result is broken, as described above
const result = parseToString(parseToDOM(templateFile));
```

### Description

This PR removes `<` from mandatory-to-replace token list for attribute parsing process.


karfau:
I thing we have multiple related issues with entities and roundtrips and this looks like a very valid (but of course breaking) fix for this specific subset of situations described.

Can you run the tests and tell jest to update the snapshots and also adopt any other failing test? Thx.

karfau:
The only thing that might be concerning is that with this change the following:
```xml
<doc a1="&quot;&lt;&amp;&gt;&apos;"></doc>
```
will be serialized to
```xml
<doc a1=\\"&quot;<&amp;>'\\"/>
```
(which is consistent with the already existing behavior for `>`)

@brodybits what do you think?

brodybits:
I wonder if we should just keep the quoted attribute value unchanged.

karfau:
> I wonder if we should just keep the quoted attribute value unchanged.

I think we should be able to add this as an option to `Node.toString`/`Document.toString()` but it would not be something that is part of the standard. And maybe this can rather happen in a different PR?

brodybits:
No more objections on my part. Thanks.

PR #174, opened 2021-01-16
Export DOMException; remove custom assertions; etc.

karfau:
and the related vows test suite that was no longer run. (see https://github.com/xmldom/xmldom/issues/72#issuecomment-761247445)

This way the test output is no longer cluttered with all those warnings.
I also took care of converting `var` to `const` and `let` in the test files I touched.


test/dom/attr.test.js
```diff
@@ -1,71 +1,61 @@
...
 	it('can properly override attribute', () => {
-		var root = new DOMParser().parseFromString(
+		const root = new DOMParser().parseFromString(
 			"<xml xmlns:a='a' xmlns:b='b' xmlns='e'><child/></xml>",
 			'text/xml'
 		).documentElement
 		root.setAttributeNS('a', 'a:a', '1')
-		assert(root.attributes.length, 4)
+		expect(root.attributes.length).toEqual(4)
 		//not standart
```
brodybits, line 48:
```suggestion
		//not standard
```

test/dom/attr.test.js
```diff
@@ -1,71 +1,61 @@
...
 
-var DOMParser = require('../../lib/dom-parser').DOMParser
-const assert = require('../assert')
-
-// Create a Test Suite
 describe('XML attrs', () => {
 	it('can properly set attribute', () => {
-		var root = new DOMParser().parseFromString('<xml/>', 'text/xml')
+		const root = new DOMParser().parseFromString('<xml/>', 'text/xml')
 			.documentElement
```
brodybits, line 8: I would reformat it like this:

```suggestion
		const root = new DOMParser()
			.parseFromString('<xml/>', 'text/xml')
			.documentElement
```

but Prettier does not give us the option: https://github.com/prettier/prettier/issues/7884 ... argh!

brodybits:
Thanks @karfau for your work on this. You should see that I have updated the title, please feel free to change it again if needed. I will likely push some updates and leave a few more suggestions.

test/dom/ns-test.test.js
```diff
@@ -15,55 +14,44 @@ describe('XML Namespace Parse', () => {
...
 		)
-		for (var _i = 0, els_1 = els; _i < els_1.length; _i++) {
-			var el = els_1[_i]
+		for (let _i = 0, els_1 = els; _i < els_1.length; _i++) {
+			const el = els_1[_i]
 
-			var te = doc.createElementNS(n1, 'test')
+			let te = doc.createElementNS(n1, 'test')
 			te.setAttributeNS(n1, 'bar', 'valx')
-			var te = doc.createElementNS(n1, 'test')
```
brodybits, line 47: I think these 2 lines are redundant and should be removed, then we can use `const` instead of `let` here.

brodybits: I have removed the 2 redundant lines here.

test/dom/attr.test.js
```diff
@@ -1,71 +1,65 @@
...
 			"<xml xmlns:a='a' xmlns:b='b' xmlns='e'><child/></xml>",
 			'text/xml'
 		).documentElement
 		root.setAttributeNS('a', 'a:a', '1')
-		assert(root.attributes.length, 4)
+		expect(root.attributes.length).toBe(4)
 		//not standart
 		//    	root.firstChild.setAttributeNode(root.attributes[0]);
-		//    	assert(root.attributes.length, 0);
+		//    	expect(root.attributes).toHaveLength(0);
```
brodybits, line 54: I think it would be really awesome if we could understand why these commented-out lines are here and if there is anything we can do to resolve them. These commented-out lines seem to have come from this commit back in 2012: c9f94a8c0c3bfd8eea5c2050bffed09d765f61f3

brodybits (commented):
LGTM if we can add `DOMException` to the documentation. I did push some updates which are mostly "nits". And a gentle reminder to merge this in a squash commit. Thanks again!

test/dom/attr.test.js
karfau: My first guess: 
It's a test for something that is not specified in WC3 standard.
If it passes we should just activate and document it (in a different PR), since people might rely on it.
If it doesn't pass we should get rid of the comment :)

Will check it.

karfau:
@brodybits  We seem to using a mix of `expect(...).toBe` and  `expect(...).toEqual` in our tests.
And it looks like you think we should use `toBe` when possible and only use `toEqual` for deep equality checks?

Should I also fix the other places of `toEqual` that compare literals? (In this PR or a separate one?)

test/dom/attr.test.js
karfau: Even though adding those lines leaves a passing test, it leaves the DOM in a broken state: the serialization of `rootNode` or `rootNode.firstChild` fails with `Attribute in use`.
Obviously it was not transferred to the new parent/owner correctly.

So I think we should just drop those lines. I don't think they hint to anything that is worth having a test for right now.

brodybits: I would kept the comment, ideally with as much historical context as we know, until we get a chance to test the functionality. As I said in <https://github.com/xmldom/xmldom/issues/72#issuecomment-761727690> this looks like untested functionality.

brodybits:
> And it looks like you think we should use `toBe` when possible and only use `toEqual` for deep equality checks?

`toBe` is preferred whenever possible (I think it is equivalent to `===`); `toStrictEqual` when comparing objects for deep equality

> Should I also fix the other places of `toEqual` that compare literals? (In this PR or a separate one?)

Yes. I think I would prefer fixing them in a separate PR.

test/dom/attr.test.js
brodybits: > Even though adding those lines leaves a passing test, it leaves the DOM in a broken state

@karfau I missed that comment, guess you are right to remove it then. Hope we get a chance to test it someday:)

lib/dom.js
```diff
@@ -1251,6 +1251,7 @@ try{
 
 //if(typeof require == 'function'){
 	exports.Node = Node;
+	exports.DOMException = DOMException;
```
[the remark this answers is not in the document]
karfau: Done

karfau:
I think I'm done with my changes in this PR. Can you have another look at the new changes?
Thx

brodybits (approved):

PR #178, opened 2021-01-17
Switch to (only) MIT license

### LINKED ISSUE #172: Conflicting license versions used for LGPL option
The [package.json](https://github.com/xmldom/xmldom/blob/master/package.json#L93) states that the LGPL license version applicable to this project is version 2.

The [LICENSE.md](https://github.com/xmldom/xmldom/blob/master/LICENSE.md) file links to the LGPL version 3 license (link leads through to [here](http://www.gnu.org/licenses/lgpl-3.0.html)).

These license versions differ in content. I can recommend using https://spdx.org/licenses/ and the short identifier to identify which license version you were intending to use.

(I'm aware that there is a choice between this and the MIT license (which is more permissive), so I'm not sure how many people will be impacted)

### LINKED ISSUE #173: MIT license is not complete
# Issue
An MIT license usually includes a copyright statement. The template which is linked to here https://github.com/xmldom/xmldom/blob/master/LICENSE.md is only a template. Most libraries I've seen use a LICENSE file with the copyright details filled in (see [restify](https://github.com/restify/node-restify/blob/master/LICENSE) as an example.

# Proposed solution
It would really help to be able to link to a file in your repo with the complete license text to make attributing and open source acknowledgements easier.

karfau:
As suggested/discussed in the following tickets:

Fixes #172
Fixes #173

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,5 @@
-Copyright 2021 Chris Brody, Tim Brust, Eric Newport and other contributors
+Copyright 2019 - present Christopher J. Brody and other contributors, as listed in: https://github.com/xmldom/xmldom/graphs/contributors
+Copyright 2012 - 2017 @jindw <jindw@xidea.org> and other contributors, as listed in: https://github.com/jindw/xmldom/graphs/contributors
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
 
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,7 @@
+Copyright 2021 Chris Brody, Tim Brust, Eric Newport and other contributors
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+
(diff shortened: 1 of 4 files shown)


LICENSE
```diff
@@ -0,0 +1,7 @@
+Copyright 2021 Chris Brody, Tim Brust, Eric Newport and other contributors
```
brodybits, line 1: - I would probably leave the year out, like this one: https://github.com/facebook/react/blob/master/LICENSE
- You forgot the original author @jindw
- I think you should add yourself first if you are planning to make more changes

karfau: Thank you for pointing to the part that I was least sure about.

> * I would probably leave the year out, like this one: https://github.com/facebook/react/blob/master/LICENSE

According to all templates I found the year is part of them (and it looks to me as if Facebook just took the liberty to drop it).
But we could say `2014 to present` to make it more maintainable.

> * You forgot the original author @jindw
> * I think you should add yourself first if you are planning to make more changes

Should we really keep a redundant list in package.json and this file?
Maybe we can rather link to the package.json file and only mention the current maintainers here?
Or list specify the years that the original author contributed?
Ideally we could mention an entity here, so it doesn't need to be updated so often, but I'm not sure xmldom org would be good enough since it's not a legal entity as far as I know.
If that's not possible I would argue for only mentioning a small amount of people (e.g. the people able to publish the NPM package) and point to the other contributors in some way.

brodybits: I think this answer can help us solve the problem with the redundant list: https://opensource.stackexchange.com/questions/4960/can-a-team-be-the-copyright-holder-mit/4970#4970

In terms of year range, I would make it "2012-present". I think it goes back to 2012 from the original author: https://github.com/xmldom/xmldom/commits/master?after=ad773c90f2e6c080ea552487b7604e9339c2a689+279&branch=master

I am also assuming that it is OK for us to "stamp" it back to 2012 even though we (you and I) did not start contributing until 2019/2020.

karfau: > I think you should add yourself first if you are planning to make more changes

Oh and I just wanted to support progress on this topic, I'm not eager right now to take more responsibility regarding xmldom than I'm already doing.

karfau:
Beside the remarks above I'm not confident when to merge this PR in general: How much approvals should we have?

LICENSE
karfau, line 1: One possible option
```suggestion
Copyright 2018 - present Chris Brody and other contributors (as listed in https://github.com/xmldom/xmldom/blob/master/package.json)
Copyright 2014 - 2017 jindw and other contributors (as listed in https://github.com/jindw/xmldom/blob/master/package.json)
```
or any subset of it

brodybits, line 1:
```suggestion
Copyright 2019 - present Christopher J. Brody and other contributors, as listed in: https://github.com/xmldom/xmldom/graphs/contributors
Copyright 2012 - 2017 @jindw <jindw@xidea.org> and other contributors, as listed in: https://github.com/jindw/xmldom/graphs/contributors
```

brodybits:
I would also be happy to take over responsibility for this, even in a separate PR if needed. I am in a webinar now, should be back in 2-3 hours. Thanks to @karfau for the much needed help over the past year.

karfau:
> I would also be happy to take over responsibility for this, even in a separate PR if needed. I am in a webinar now, should be back in 2-3 hours. 

Not sure what this relates to.
I would prefer that you are deciding when to land this PR and not wait for me to do that.

> Thanks to @karfau for the much needed help over the past year.

You're very welcome :)

PR #180, opened 2021-01-19
docs: Configure `auto-changelog` for generating updated CHANGELOG.md

karfau:
- added script `changelog` to add "Unreleased" section to changelog, so it can be tweaked manually
- `auto-changelog` is invoked using npx, so it's not a `devDependency`

https://github.com/CookPete/auto-changelog

PS: This is just #138 without any (real) changes to CHANGELOG.md

### DIFF
--- a/auto-changelog.hbs
+++ b/auto-changelog.hbs
@@ -1,0 +1,34 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+{{#each releases}}
+  {{#if href}}
+    ## [{{title}}]({{href}}){{#if tag}} - {{isoDate}}{{/if}}
+  {{else}}
+    ## {{title}}{{#if tag}} - {{isoDate}}{{/if}}
+  {{/if}}
+
+  {{#if summary}}
+    {{summary}}
+  {{/if}}
+
+  {{#if merges}}
+    ### Merged
+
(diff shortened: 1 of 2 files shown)


karfau:
FYI:
- This is what it looks like when running `npm run changelog` right now: https://github.com/karfau/xmldom/commit/e288aa7d4947044578bed6d92c505a64b7288e0b I'm not adding this to the PR on purpose. It's hard to maintain and not part of the goal of this PR.
- In the mean time I have collected some experience with automated releases that happen whenever they are required based on https://semantic-release.gitbook.io/semantic-release/ but it works based on (conventional commits](https://conventionalcommits.org/) that we would need to "enforce", especially when merging PRs. In my personal perspective it's the way to go to get rid of the manual burden of doing releases.
- There are other alternatives where I'm not aware of the details:
  - https://github.com/atlassian/changesets
  - https://github.com/nodejs/changelog-maker
  - ???
  And there are many more options that are based on commit message conventions.

If it's better suited we can close this PR and first discuss it in a Meta issue.

brodybits:
@karfau it would be awesome if you could raise meta issues for both the auto changelog and the semantic release.

These are definitely new for me and I think I need to catch up with these tools.

I definitely recall seeing the Conventional Commits and semantic-release tool in the past, but I think this is the first place I have seen the `auto-changelog`.

In general I would downvote including anything related to test or devDependency updates in the changelog. But I would be open if there are any good arguments for including these items.

Issue #58, opened 2020-06-24
Some entities in text content are not converted/escaped when serializing


cburatto:
I have noticed that node values/strings containing `>` are not converted to `&gt;` when serializing the DOM. Is this due to any flag I should be using?

karfau:
Did is the second > in your question written as `&gt;`? (Surround it with back ticks to prevent github from rendering it as >.)
If so:
I'm not aware of any "flags" that would change behavior of handling `>` in those cases.
I'm aware that xmldom is not doing that currently but (because I'm working on [restoring the tests](#35 )) I'm confident that the parsing is still correctly done.

I think having something that looks like a closing tag bracket in text or values is "uncritical" for XML parsers, which can not be said about `<` (these need to be converted for the parsing to work).

I have also seen that other parsers (at least one of the abandoned domjs or xmljs) treat those things in a more consistent manner.

If we would decide to change that (after having a stable test suite running on every change), even if it's not really a difference (from my current understanding) it might be considered a breaking change.

So my guess is such a change might not come very soon.

cburatto:
Thanks -- I do mean a tag bracket character. For example, if a node value contains HTML text where entities have been converted. You load it with the DOMParser, then when you XMLSerialize it the entities are reconverted, except the closing bracket `>`

Example: 
`<xmlelement>&lt;b&gt;This is bold&lt;/b&gt;</xmlelement>`

gets serialized as
`<xmlelement>&lt;b>This is bold&lt;/b></xmlelement>`

karfau:
Yes. That's what I was also referring to. But another short search in the old repo shows that there are already multiple issues around it there and also PRs.
I linked the "main thread" above.

karfau:
@sarod Do I understand you correctly that you agree to a "subset" of the original question, namely for the case of `]]>` that needs to be converted to `]]%gt;`?

Do you think we should just always convert it, even thought he specification you are quoting also says (emphasis mine)

> The right angle bracket (>) **may** be represented using the string " &gt; ", and must, for compatibility, be escaped using either " &gt; " or a character reference when it appears in the string " ]]> " in content, when that string is not marking the end of a CDATA section.

?

I think I'm convinced that we should take care of the `]]>` case.
But for anything else I would love to see a failing test (either in a comment or as a PR or as link to any repo).

karfau:
[...]

PR #181, opened 2021-01-19
fix(dom): Escape `]]>` when serializing CharData

karfau:
to adhere to the XML specification

https://www.w3.org/TR/xml/#NT-CharData
https://github.com/xmldom/xmldom/issues/58#issuecomment-743231935

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -402,12 +402,17 @@ Node.prototype = {
     }
 };
 
+const AMP = '&amp;';
+const GT = '&gt;';
+const LT = '&lt;';
+const CDATA_END_ESCAPED = ']]'+GT;
+const QUOT = '&quot;';
 
 function _xmlEncoder(c){
-	return c == '<' && '&lt;' ||
-         c == '>' && '&gt;' ||
-         c == '&' && '&amp;' ||
-         c == '"' && '&quot;' ||
+	return c == '<' && LT ||
+         c == '>' && GT ||
+         c == '&' && AMP ||
+         c == '"' && QUOT ||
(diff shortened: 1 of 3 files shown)


lib/dom.js
```diff
@@ -1066,7 +1066,25 @@ function serializeToString(node,buf,isHTML,nodeFilter,visibleNamespaces){
...
+		 * when that string is not marking the end of a CDATA section.
+		 *
+		 * In the content of elements, character data is any string of characters
+		 * which does not contain the start-delimiter of any markup
+		 * and does not include the CDATA-section-close delimiter, `]]>`.
+		 *
+		 * @see https://www.w3.org/TR/xml/#NT-CharData
+		 */
+		return buf.push(node.data
+			.replace(/[<&]/g,_xmlEncoder)
```
brodybits, line 1085: I wonder if this may be better performant or not:

```suggestion
			.replace(/&/g, '&amp;')
			.replace(/</g, '&lt;')
```

and consider replacing hard-coded patterns & strings with `const` values

brodybits:
LGTM in general. Unfortunately some of this is very new for me and I am relying on the comments and my understanding of the referenced specs. Thanks for working on this.

lib/dom.js
karfau: I extracted some `const`s, but I don't think the performance difference between the two regex approaches is a relevant/important enough concern in this case.

karfau:
I needed to revert the usage of `const` since we decided to stick to ES5 syntax in `lib`. Thx eslint for reminding me.

PR #183, opened 2021-01-25
Improve some XML test strings

### LINKED ISSUE #128: Odds & ends from Prettier formatting of test code [PR #123]
@karfau identified quite a few things in review of PR #123 that should be resolved.

brodybits:
resolves #128 (followup to PR #123 - reformat test code using Prettier)

### DIFF
--- a/test/dom/element.test.js
+++ b/test/dom/element.test.js
@@ -51,10 +51,12 @@ describe('XML Namespace Parse', () => {
 
 	it('supports getElementsByTagNameNS', () => {
 		const doc = new DOMParser().parseFromString(
-			'<xml xmlns="http://test.com" xmlns:t="http://test.com" xmlns:t2="http://test2.com">' +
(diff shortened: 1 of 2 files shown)


karfau (approved):
LGTM
(Even though I'm not aware why this way of concatenating strings is any different from the other way.)

brodybits:
Better alignment. Yeah maybe not really worth it?

karfau:
> Better alignment. Yeah maybe not really worth it?

Did you already handle all of them?
Since you invested the time we can also land it. But I'm not sure we will be able to keep it this way in the future...
Your choice :)

karfau:
Since you didn't land it: Couldn't/Shouldn't we nowadays just use template strings everywhere, consistently?

brodybits:
> Shouldn't we nowadays just use template strings everywhere, consistently?

Makes sense to me, yes.

PR #195, opened 2021-03-09
refactor: move all require statements to beginning of lib/dom-parser.js

brodybits:


lib/dom-parser.js
```diff
@@ -1,3 +1,10 @@
+var htmlEntity = require('./entities');
+var sax = require('./sax');
+var XMLReader = sax.XMLReader;
+var ParseError = sax.ParseError;
+var DOMImplementation = exports.DOMImplementation = require('./dom').DOMImplementation;
+exports.XMLSerializer = require('./dom').XMLSerializer ;
```
brodybits, line 6: TODO:

- [ ] move exports to the end of this module
- [ ] I think we should add a separate line such as `var dom = require('./dom')` rather than importing it twice.

brodybits:
Superseded by PR #233

PR #199, opened 2021-03-13
fix: Escape `<` when serializing attribute values

### LINKED ISSUE #198: '<' and '>' are not escaped in attribute values
Serializer does not escape < and > characters in attribute values
`const doc = new DOMParser().parseFromString('<a a="&lt;&gt;"/>');`
`const result = new XMLSerializer().serializeToString(doc);`
result is `<a a="<>"/>`, which is invalid xml

karfau:
to produce well formed XML.

> Well-formedness constraint: No `<` in Attribute Values
> The replacement text of any entity referred to directly or indirectly in an attribute value must not contain a `<`.

https://www.w3.org/TR/xml/#CleanAttrVals
https://www.w3.org/TR/xml/#NT-AttValue

fixes #198

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -1068,7 +1068,13 @@ function serializeToString(node,buf,isHTML,nodeFilter,visibleNamespaces){
 		}
 		return;
 	case ATTRIBUTE_NODE:
-		return buf.push(' ',node.name,'="',node.value.replace(/[&"]/g,_xmlEncoder),'"');
+		/**
+		 * Well-formedness constraint: No < in Attribute Values
+		 * The replacement text of any entity referred to directly or indirectly in an attribute value must not contain a <.
+		 * @see https://www.w3.org/TR/xml/#CleanAttrVals
+		 * @see https://www.w3.org/TR/xml/#NT-AttValue
+		 */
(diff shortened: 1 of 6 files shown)


lib/dom.js
```diff
@@ -1068,7 +1068,13 @@ function serializeToString(node,buf,isHTML,nodeFilter,visibleNamespaces){
...
 		return;
 	case ATTRIBUTE_NODE:
-		return buf.push(' ',node.name,'="',node.value.replace(/[&"]/g,_xmlEncoder),'"');
+		/**
+		 * Well-formedness constraint: No < in Attribute Values
+		 * The replacement text of any entity referred to directly or indirectly in an attribute value must not contain a <.
+		 * @see https://www.w3.org/TR/xml/#CleanAttrVals
+		 * @see https://www.w3.org/TR/xml/#NT-AttValue
+		 */
+		return buf.push(' ',node.name,'="',node.value.replace(/[<&"]/g,_xmlEncoder),'"');
```
brodybits, line 1077: minor nit, just a suggestion:

```suggestion
		return buf.push(' ', node.name, '="', node.value.replace(/[<&"]/g,_xmlEncoder), '"');
```

brodybits (approved):

PR #201, opened 2021-03-21
test(stryker): Replace line numbers by error index

### LINKED ISSUE #200: Stryker test failures on GitHub
Quoting @karfau from <https://github.com/xmldom/xmldom/pull/192#issuecomment-798921808>:

> the new tests added as part of the security fix break the stryker setup, since they rely on line numbers in snapshots, and stryker mutates the files causing line numbers to change

---

P.S. I am wondering if the most recent Stryker update would suffer from this issue or not.

karfau:
so the snapshots match even when stryker mutates the files.

The intermediate mapping is also stored to `tests/error/reported.json` for easier inspection, so I `.gitignore`d it.

The good news: The score improved quite a bit :tada: 
image

Fixes #200


brodybits:
Thanks for your timely work on this.

My one, minor question is that it looks like this is using a possibly hackish utility function, with "sax.js" hardcoded multiple times. And I am not sure if I understand it 100%. I am thinking it may be ideal if we can find a better way to improve the reusability.

karfau:
I think I should be able to do another round of self review and try to refactor/document some more things to make them more generic & easy to understand. 
But I'm not sure how fast I'll be able to do that.

Of course the whole test suite about the reported errors and warnings is not trivial to understand.

brodybits:
> Of course the whole test suite about the reported errors and warnings is not trivial to understand.

I think that is the key, and to be extra clear I think this is the result of the original design.

Further refactoring and improvement would be awesome, assuming you have the time for it. Otherwise I would not object to improving it someday later. Thanks again for all of your care and efforts on such a messy code base!

karfau:
@brodybits I'm not sure what to understand from your reply. 
Do you think we can land it as is to solve the stryker issue and improve it later, or do do you prefer we invest more time into this before we land it?
(From my perspective, we should land it so we can take the next steps towards 0.6.0 .)

brodybits:
It is fine with me if you want to land these changes now.

While I think it would be nice to land these changes in a better state, it would probably be better to fix the build now.

Thanks again.

karfau:
@brodybits I think I did everything I could to improve it in this short period of time.

Please have a look, if this is the direction you were talking about.
You can land the PR or I will do it in some days, if you don't give me a different signal.

brodybits:
From a quick read it looks better to me, no more objections on my part. I would like to leave this up to @karfau to merge.

I am also very happy to see the improvement in the mutation score.

PR #206, opened 2021-03-30
ci: run Stryker on Node.js 14

brodybits:
Stryker seems to take 2 hours on GitHub, maybe a little longer with the recent update. I wonder if it may go faster on Node.js 14.

### DIFF
--- a/.github/workflows/stryker.yml
+++ b/.github/workflows/stryker.yml
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


karfau (approved):
Not sure why this would make a difference, but I'm fine with trying it.

brodybits:
When I tried with Node.js 12 vs 14 with 2 vCPU, 2 test runners which is similar to what I see on GitHub, I noticed very little difference in elapsed time.

PR #207, opened 2021-03-30
ci: Node.js versions on multiple lines

brodybits:


.github/workflows/test-node.js.yml
```diff
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 14.x]
+        node-version: [10.x, 12.x, 14.x, 15.x]
```
brodybits, line 19: I think we should split this into multiple lines, like this for example
```suggestion
        node-version:
        - 10.x
        - 12.x
        - 14.x
        - 15.x
```

karfau (approved):

.github/workflows/test-node.js.yml
```diff
@@ -16,7 +16,11 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 14.x]
+        node-version:
+        - 10.x
+        - 12.x
+        - 14.x
+        - 15.x
```
brodybits, line 23: I am now thinking we can leave Node.js 15 out since initial release of Node.js 16 should be available in the next few days: https://nodejs.org/en/about/releases/

```suggestion
```

karfau: So let's add 16 when it's ready?

brodybits: Yes. Should I wait for 16 before merging this one?

karfau:
> Yes. Should I wait for 16 before merging this one?

I'm fine with either approach, so far this PR only contains a formatting change, so if you want to land that, just go ahead.

karfau:
v16 has been released

PR #208, opened 2021-03-31
test: Stryker Jest runner with config workarounds

brodybits:
__updated:__ I reconfigured by removing `stryker.conf.json` from my work area, running `npx stryker init` again, and re-patching `stryker.conf.json` with some of the configuration from the master branch. The reconfiguration gave me the `coverageAnalysis` entry changed to `perTest`.

I was hoping to improve the speed on GitHub. I tried with and without this update on cloud instances with 2 vCPU, for 2 test runner processes, which seems to reflect what happens on GitHub. But it took about half minute longer with the Jest runner, which is insignificant considering that it takes almost 2 hours on GitHub with 2 test runner processes.

I had to update `jest.config.js` to configure `testMatch` instead of `testRegex`. From the debug tracing command, as directed when Stryker fails, it looked to me like with the Jest runner it was messing up the configuration of `testRegex`. I suspect that this is why the jest-runner did not work in PR #140 (update to Stryker 4). I think it would be ideal if we could get this fixed on Stryker at some point. My changes may need cleanup, and I am not sure if we need to configure `testPathIgnorePatterns` or not.

Test time was almost the same, maybe half a minute longer with the Jest runner. See <https://github.com/stryker-mutator/stryker-js/issues/1514#issuecomment-675571410>:

> Mutation switching _by itself_ will not have a significant impact on Jest projects unfortunately. You might have some performance improvement because Jest would be able to cache build results, but definitely not an order of magnitude. However, it does open up the way for coverage analysis in jest. Hot reload should theoretically also be possible with Jest, but not with their current public api alone. We'll be focussing on jest next, after we've released v4.

### End results with command runner

```console
Mutation testing  [==================================================] 100% (elapsed: ~1h 53m, remaining: n/a) 2014/2014 tested (695 survived, 41 timed out)
[...]
Ran 0.98 tests per mutant on average.
---------------|---------|----------|-----------|------------|----------|---------|
File           | % score | # killed | # timeout | # survived | # no cov | # error |
---------------|---------|----------|-----------|------------|----------|---------|
All files      |   65.49 |     1278 |        41 |        695 |        0 |       0 |
 dom-parser.js |   80.00 |      156 |         0 |         39 |        0 |       0 |
 dom.js        |   65.81 |      590 |        24 |        319 |        0 |       0 |
 entities.js   |    7.85 |       19 |         0 |        223 |        0 |       0 |
 sax.js        |   82.30 |      513 |        17 |        114 |        0 |       0 |
---------------|---------|----------|-----------|------------|----------|---------|
[...]
23:51:59 (1131) INFO MutationTestExecutor Done in 113 minutes 37 seconds.
```

### End results with Jest runner

```console
Mutation testing  [==================================================] 100% (elapsed: ~1h 53m, remaining: n/a) 1832/1832 tested (513 survived, 92 timed out)
[...]
Ran 183.07 tests per mutant on average.
---------------|---------|----------|-----------|------------|----------|---------|
File           | % score | # killed | # timeout | # survived | # no cov | # error |
---------------|---------|----------|-----------|------------|----------|---------|
All files      |   65.42 |     1223 |        92 |        513 |      182 |       4 |
 dom-parser.js |   80.00 |      149 |         7 |         24 |       15 |       0 |
 dom.js        |   65.81 |      585 |        29 |        168 |      151 |       0 |
 entities.js   |    7.85 |       19 |         0 |        223 |        0 |       0 |
 sax.js        |   82.19 |      470 |        56 |         98 |       16 |       4 |
---------------|---------|----------|-----------|------------|----------|---------|
[...]
23:55:19 (1135) INFO MutationTestExecutor Done in 114 minutes 11 seconds.
```

I would also like to investigate the 4 errors on `sax.js` at some point.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -143,21 +143,24 @@ module.exports = {
   // Adds a location field to test results
   // testLocationInResults: false,
 
+  // TBD ENABLED to work with Stryker Jest runner:
   // The glob patterns Jest uses to detect test files
-  // testMatch: [
-  //   "**/__tests__/**/*.[jt]s?(x)",
-  //   "**/?(*.)+(spec|test).[tj]s?(x)"
-  // ],
+  testMatch: [
+    "**/__tests__/**/*.[jt]s?(x)",
+    "**/?(*.)+(spec|test).[tj]s?(x)"
+  ],
 
   // An array of regexp pattern strings that are matched against all test paths, matched tests are skipped
-  // testPathIgnorePatterns: [
+  testPathIgnorePatterns: [
   //   "/node_modules/"
-  // ],
+    "/test/test.js"
+  ],
 
+  // TBD NOT WORKING with Stryker Jest runner:
   // The regexp pattern or array of patterns that Jest uses to detect test files
-  testRegex: [
-  	/\.test\.js?$/
-	],
+  //  testRegex: [
+  //    /\.test\.js?$/
+  //  ],
 
   // This option allows the use of a custom results processor
   // testResultsProcessor: undefined,
--- a/stryker.conf.json
+++ b/stryker.conf.json
@@ -7,14 +7,11 @@
     "progress",
     "dashboard"
   ],
-  "testRunner": "command",
+  "testRunner": "jest",
+  "coverageAnalysis": "perTest",
   "dashboard": {
     "project": "github.com/brodybits/xmldom",
     "version": "master"
-  },
-  "coverageAnalysis": "off",
-  "commandRunner": {
-    "command": "npm run test:jest -- --bail --no-cache --ci --maxWorkers=1"
   },
   "timeoutMS_comment": "When all tests pass (mutant survives) our tests need some time. So this is configured to prevent 'timeout's",
   "timeoutMS": 15000
--- a/package-lock.json
+++ b/package-lock.json
@@ -864,6 +864,18 @@
         "weapon-regex": "~0.4.1"
       }
     },
+    "@stryker-mutator/jest-runner": {
+      "version": "4.5.1",
+      "resolved": "https://registry.npmjs.org/@stryker-mutator/jest-runner/-/jest-runner-4.5.1.tgz",
+      "integrity": "sha512-ei0YWdf3tjShChvU5cBBET3Eqa3PfshcfHxAG/RgYgrfZYGHL1GnT5dsGLojf4b5TEPY12Q9HVkSnsnoFvW5mQ==",
+      "dev": true,
+      "requires": {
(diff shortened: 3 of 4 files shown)


stryker.conf.json
```diff
@@ -7,15 +7,12 @@
     "progress",
     "dashboard"
   ],
-  "testRunner": "command",
+  "testRunner": "jest",
+  "coverageAnalysis": "perTest",
```
brodybits, line 11: I would like to try changing `coverageAnalysis` back:

```suggestion
  "coverageAnalysis": "off",
```

When I tried this, it seemed to be testing 2014 mutations again.

jest.config.js
```diff
@@ -144,20 +144,22 @@ module.exports = {
...
+  ],
 
   // An array of regexp pattern strings that are matched against all test paths, matched tests are skipped
-  // testPathIgnorePatterns: [
-  //   "/node_modules/"
-  // ],
+  testPathIgnorePatterns: [
+   "/test/test.js",
+    "/node_modules/"
+  ],
```
brodybits, line 156: I am not sure at this point whether or not this `testPathIgnorePatterns` entry is needed, still need to test to be sure.

brodybits: Commenting this out causes the initial test run to fail with Stryker 4.5.1, will keep for now, at least.

brodybits:
As I said in PR #220 Stryker 4.6.0 update seems to run much slower, proposing to use Jest runner with Stryker 4.5.1 for now at least.

stryker.conf.json
karfau: I would have thought/expected, that this would be the most important part of speeding up the mutation testing, since it only needs to execute those tests for any mutation, that would be affected by it.

brodybits: Sounds right *in theory*. When I tried `off` vs `perTest` vs `all`, both `perTest` and `all` resulted in 1800+ mutations while `off` resulted in 2014 mutations (same as command runner). I recall that the initial focus of Stryker 4 was the mutation switching and that improving the Jest runner would be done sometime as a followup.

My primary purpose of this proposal was to be ready for the future, assuming that they will make some improvements in the Jest runner.

brodybits:
In terms of status: I have tested this with Stryker 4, not sure if it will work with Stryker 5 or not. I am also not sure whether or not these workarounds are needed with Stryker 5.x as well.

In case we do still need these workarounds with Stryker 5.x, I think it would be good to investigate *why*.

karfau:
We are no longer using stryker mutation testing, so this PR is obsolete

Issue #203, opened 2021-03-30
`&amp;` parsing wrong when `xmlns` points to XHTML


weiwu-zhang:
````
$ cat xmldom_bug_22.js 
const { DOMParser } = require('@xmldom/xmldom')

const items = [
  `<html xmlns="http://www.w3.org/1999/xhtml"><script>let message = " &amp; ETH";</script></html>`,
  `<html><script>let message = " &amp; ETH";</script></html>`,
];

for (const item of items) {
  const xml = new DOMParser().parseFromString(item, "application/xml");
  console.log(xml.documentElement.childNodes[0].textContent);
}
$ node xmldom_bug_22.js 
let message = " &amp; ETH";
let message = " & ETH";
````

Where both messages should be the same. Something trigged the misintrepreation when `xmlns="http://www.w3.org/1999/xhtml"` presents.

P.S. this is not a duplicate of #58

karfau:
@colourful-land Thx for reporting this, providing a test scenario and linking all related issues.

1. Just to be sure: What versions of xmldom have you tried this with?

2. Did I understand correctly, that you expect the output to be `let message = " & ETH";` in both cases?

colourful-land:
>     1. Just to be sure: What versions of xmldom have you tried this with?

I first tried 0.5.0 - failing the test, I tried again with the master branch from github, still fails.

>     2. Did I understand correctly, that you expect the output to be `let message = " & ETH";` in both cases?

Yes, as it is the case if the code runs in a browser using the browser's dom parser.

karfau:
FYI: I have a rough plan how to fix this issue (I already invested quite some time into a draft so far), but it's not a trivial one, and it's a breaking change.

The main issue is that xmldom currently "detects" the default HTML namespace when parsing and applies certain rules, that should **only** be applied when setting the second argument `mimeType` of `DOMParser.parseFromString()` to `'text/html'`. (According to [the most current spec](https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#dom-domparser-parsefromstring).)

Since there will be quite some changes and quite some missing tests for the areas of code that I need to touch, I will file multiple PRs before we can resolve this issue.

colourful-land:
Thank you very much @karfau. Right now I simply replace all occurrences of xhtml namespace in lib/sax.js and lib/dom.js to keep my project going. Hope to see a new release so I can rever to use unmodified xmldom in the near future!

karfau:
[...]

Issue #191, opened 2021-03-06
Suggestion: Add typescript types to package


karfau:
There are some issues with the typescript types provided by `@types/xmldom`, that are hard to resolve, since they rely on the DOM types provided by typescript:
https://github.com/DefinitelyTyped/DefinitelyTyped/issues/49953

I think the easiest way to resolve that issue and maintain the types is to add doc comments to the code in `lib` and generate the types from it.

The main advantage is also that the types don't claim to support APIs that it actually doesn't support, which currently would lead to a runtime error.

brodybits:
I think this sounds good. We will probably want to ask DefinitelyTyped to deprecate the `@types/xmldom` package. It would also be ideal to add tests for the TypeScript types, as I have seen in some packages by Sindre Sorhus.

kachkaev:
Some thoughts on this following a recent change in library name: https://github.com/xmldom/xmldom/issues/271#issuecomment-902217165

karfau:
Thx to @kachkaev for pointing out that since we resolved #271 by changing our package name to `@xmldom/xmldom` the types provided by `@types/xmldom` no longer work.

Personally I would favour to generate the type definitions from our source, even if they are not yet ideal. Any support is welcome.

kachkaev:
@karfau I agree with you that providing the typings from source is the best solution. It minimizes the probability of discrepancies between the declarations and the implementation. It might take some time to setup though depending on the current state of the codebase, leaving folks with no typings for some time.

If you generally support the migration of typings into the library, how about doing this in two steps?

1. Manually add existing typings from [DefinitelyTyped → index.d.ts](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/xmldom/index.d.ts)
   --Release a patch version--
2. Change how the lib is built, thus providing 'true' typings from source
   --Release a patch or a minor version--

Happy to help with 1 at least.

brodybits:
An alternative idea could be to first add the types and then add a typecheck step to verify the consistency. This could help us avoid the complexities and limitations of adding a build pipeline, making it easier to install directly from GitHub, for example.

By "typecheck step" it should be possible to use the TypeScript compiler to check consistency between types in the DTS and types in our JavaScript code with help from JSDoc comments, like I started contributing to Prettier.

karfau:
@kachkaev 

I disagree that we need to change how the lib s build to generate types. they can also be generated from commented JS files.

[...]

PR #209, opened 2021-04-03
refactor: Extract immutable conventions & entities, etc.

karfau:
Before starting to work on fixing #203, where namespaces and mime types are going to be used in multiple places, I decided to gather them in a separate module.

**conventions.js**
- `freeze`: is a small wrapper around `Object.freeze`, that gives type hints
  and also "works" if `Object.freeze` should not be available at runtime
- `MIME_TYPE`: has all mime types that are supported by `DOMParser.parseFromString` (second argument)
- `MIME_TYPE.isHTML`: short way to check if a mime type is the only one that enables HTML parsing
- `NAMESPACE`: has all namespaces that we are currently using in the code base
- `NAMESPACE.isHTML`: to check if a given value is the XHTML namespace

**entities.js**
- deprecated `entityMap` (but it's still available as an alias to `HTML_ENTITIES`)
- `HTML_ENTITIES`: new name for `entityMap`, is now immutable
- `XML_ENTITIES`: immutable map for the predefined XML entities

### DIFF
--- a/lib/conventions.js
+++ b/lib/conventions.js
@@ -40,6 +40,7 @@ var MIME_TYPE = freeze({
 	 * @see https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#dom-domparser-parsefromstring WHATWG HTML Spec
 	 */
 	HTML: 'text/html',
+
 	/**
 	 * Helper method to check a mime type if it indicates an HTML document
 	 *
@@ -53,6 +54,7 @@ var MIME_TYPE = freeze({
 	isHTML: function (value) {
 		return value === MIME_TYPE.HTML
 	},
+
 	/**
 	 * `application/xml`, the standard mime type for XML documents.
 	 *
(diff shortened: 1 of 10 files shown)


brodybits:
Thanks @karfau, definitely looks really nice in general. I will try to review this more carefully later this week or sometime next week, please ping me in case it slips.

Do we want to get this or any other recent PRs into the 0.6.0 release?

If not, I would favor that we publish the 0.6.0 release before merging any of these recent PRs.

lib/conventions.js
```diff
@@ -0,0 +1,144 @@
...
+ *
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString#Argument02 MDN
+ * @see https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#domparsersupportedtype WHATWG HTML Spec
+ * @see DOMParser.prototype.parseFromString
+ */
+var MIME_TYPE = freeze({
+	/**
+	 * `text/html`, the only mime type that triggers treating an XML document as HTML.
+	 *
+	 * @see DOMParser.SupportedType.isHTML
```
brodybits, line 36: It would be interesting to see how cross-references like this one show up when generating JSDoc pages.

brodybits (approved):

lib/conventions.js
karfau: Just did a small experiment using typedoc, which requires typescript which requires adding a `tsconfig.json` ... so I think we need to put it as an extra issue that comes right after #191

PR #210, opened 2021-04-04
feat: Update `DOMImplementation` according to recent specs

karfau:
- deprecate `hasFeature` and simplify implementation according to the latest spec
- `createDocumentType` sets `publicId` and `systemId` to an empty string instead of `undefined`
- `createDocument` sets `docType` to `null` instead of `undefined`
- export `DocumentType`, `Element`, `Node` and `NodeList` to verify `instanceof` in tests
- update doc comments with links to specs and type information

This PR is part of the preparation for solving #203

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -287,7 +287,7 @@ DOMImplementation.prototype = {
 	 * @deprecated It is deprecated and modern browsers return true in all cases.
 	 *
 	 * @param {string} feature
-	 * @param {string=} version
+	 * @param {string} [version]
 	 * @returns {boolean} always true
 	 *
 	 * @see https://developer.mozilla.org/en-US/docs/Web/API/DOMImplementation/hasFeature MDN
@@ -341,8 +341,8 @@ DOMImplementation.prototype = {
 	 *   (when parsing XML strings, the SAX parser takes care of that)
 	 *
 	 * @param {string} qualifiedName
-	 * @param {string=} publicId
-	 * @param {string=} systemId
+	 * @param {string} [publicId]
+	 * @param {string} [systemId]
 	 * @returns {DocumentType} which can either be used with `DOMImplementation.createDocument` upon document creation
 	 * 				  or can be put into the document via methods like `Node.insertBefore()` or `Node.replaceChild()`
 	 *
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -246,55 +246,108 @@ NamedNodeMap.prototype = {
 		return null;
 	}
 };
+
 /**
- * @see http://www.w3.org/TR/REC-DOM-Level-1/level-one-core.html#ID-102161490
+ * The DOMImplementation interface represents an object providing methods
+ * which are not dependent on any particular document.
+ * Such an object is returned by the `Document.implementation` property.
+ *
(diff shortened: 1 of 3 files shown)


lib/dom.js
```diff
@@ -246,55 +246,108 @@ NamedNodeMap.prototype = {
...
+ * @see https://dom.spec.whatwg.org/#domimplementation DOM Living Standard
  */
-function DOMImplementation(/* Object */ features) {
-	this._features = {};
-	if (features) {
-		for (var feature in features) {
-			 this._features = features[feature];
-		}
-	}
-};
```
karfau, line 259: The previous implementation left the definition of features and versions that are supported to the code that calls the constructor, which doesn't make any sense at all. This was one more reason why I decided to not keep or fix it, but to implement it as it is specified in the living standard.

lib/dom.js
```diff
@@ -246,55 +246,108 @@ NamedNodeMap.prototype = {
...
-		if (versions && (!version || version in versions)) {
+	/**
+	 * The DOMImplementation.hasFeature() method returns a Boolean flag indicating if a given feature is supported.
+	 * The different implementations fairly diverged in what kind of features were reported.
+	 * The latest version of the spec settled to force this method to always return true, where the functionality was accurate and in use.
+	 *
+	 * @deprecated It is deprecated and modern browsers return true in all cases.
+	 *
+	 * @param {string} feature
+	 * @param {string=} version
```
brodybits, line 277: This looks like Google Closure. Should we consider JSDoc typing:
```suggestion
	 * @param {string} [version]
```

?

lib/dom.js
```diff
@@ -246,55 +246,108 @@ NamedNodeMap.prototype = {
...
+	/**
+	 * Returns a doctype, with the given `qualifiedName`, `publicId`, and `systemId`.
+	 *
+	 * __This behavior is slightly different from the in the specs__:
+	 * - this implementation is not validating names or qualified names
+	 *   (when parsing XML strings, the SAX parser takes care of that)
+	 *
+	 * @param {string} qualifiedName
+	 * @param {string=} publicId
+	 * @param {string=} systemId
```
brodybits, line 332:
```suggestion
	 * @param {string} [publicId]
	 * @param {string} [systemId]
```

??

test/dom/dom-implementation.test.js
```diff
@@ -0,0 +1,165 @@
+'use strict'
+
+const {
+	DocumentType,
+	DOMImplementation,
+	Element,
+	Node,
+	NodeList,
+} = require('../../lib/dom')
```
brodybits, line 9: I would favor that we export and use these from `lib/index.js`, as I proposed in PR #233.

lib/dom.js
karfau: From the perspective of typescript both ways work:
https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#param-and-returns
and I perceived it easier to understand (and remember), since a parameter being optional for me is more related to the type, not the name of the field.

But I don't care. So I changed all the places.

test/dom/dom-implementation.test.js
karfau: Does it mean you want to land 233 first?

PR #213, opened 2021-04-08
fix(dom): Implement `getElementsByClassName` as specified

ChALkeR:
Reimplements `Document.getElementsByClassName` without using a dynamically generated regex.

1. Unlike _CSS grammar_, there seems to be no limitation on which classnames are valid in HTML/DOM (given that they don't contain space chars). That is, `[` seems to be a valid classname, which was broken in this impl as it was fed into a regexp unescaped and caused `xmldom` to throw.
2. Whitespace processing should be now correct. `\s` in JS regexps includes more chars than [space-separated tokens](https://www.w3.org/TR/html52/infrastructure.html#set-of-space-separated-tokens) of html.
3. This now should support a list of space-separated classnames as an argument. Previously, that was broken and order-dependent.
4. This prevents causing a ReDoS by passing in exponential regexps as strings to `getElementsByClassName()` method,
    e.g. `new RegExp("(^|\\s)" + '(((a||||)+)+)+' + "(\\s|$)").test('aaaaab')`.

Refs:
1. https://dom.spec.whatwg.org/#dom-document-getelementsbyclassname
2. https://dom.spec.whatwg.org/#dom-element-getelementsbyclassname
3. https://dom.spec.whatwg.org/#concept-getelementsbyclassname
4. https://www.w3.org/TR/html52/dom.html#element-attrdef-global-class
5. https://www.w3.org/TR/html52/infrastructure.html#set-of-space-separated-tokens
6. https://github.com/xmldom/xmldom/issues/24
7. https://github.com/xmldom/xmldom/pull/25

Note: I couldn't find the tests for this and am unsure how to test this. cc @codler perhaps — could you confirm this would work?
Seems to work for me with simple playground tests.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -675,6 +675,19 @@ Document.prototype = {
 		return rtv;
 	},
 	
+	/**
+	 * The `getElementsByClassName` method of `Document` interface returns an array-like object 
+	 * of all child elements which have **all** of the given class name(s).
+	 * 
+	 * Warning: This is a live LiveNodeList.
+	 * Changes in the DOM will reflect in the array as the changes occur.
+	 * If an element selected by this array no longer qualifies for the selector,
+	 * it will automatically be removed. Be aware of this for iteration purposes.
+	 *
+	 * @param {string} classNames is a string representing the class name(s) to match; multiple class names are separated by (ASCII-)whitespace
+	 *
+	 * @see https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementsByClassName
+	 */
 	getElementsByClassName: function(classNames) {
 		var spaceSeparatedTokens = function(string) {
 			// 5 space characters, as defined in html spec
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -22,6 +22,19 @@ function splitOnASCIIWhitespace(input) {
 	return input.split(/[\t\n\f\r ]+/).filter(notEmptyString)
 }
 
+/**
+ * Uses `list.indexOf` to implement something like `Array.prototype.includes`,
+ * which we can not rely on being available.
+ *
+ * @param {any[]} list
+ * @returns {function(any): boolean}
+ */
+function arrayIncludes (list) {
+	return function(element) {
+		return list && list.indexOf(element) !== -1;
+	}
+}
+
 function copy(src,dest){
 	for(var p in src){
 		dest[p] = src[p];
@@ -726,9 +739,7 @@ Document.prototype = {
 							var matches = classNames === classActual;
 							if (!matches) {
 								var namesActual = splitOnASCIIWhitespace(classActual)
-								matches = namesTarget.every(function(name) {
-									return namesActual.indexOf(name) !== -1
-								})
+								matches = namesTarget.every(arrayIncludes(namesActual))
 							}
 							if(matches) {
 								ls.push(node);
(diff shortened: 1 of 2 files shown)


lib/dom.js
```diff
@@ -606,14 +606,27 @@ Document.prototype = {
...
 	
-	getElementsByClassName: function(className) {
-		var pattern = new RegExp("(^|\\s)" + className + "(\\s|$)");
+	getElementsByClassName: function(classNames) {
+		var spaceSeparatedTokens = function(string) {
+			// 5 space characters, as defined in html spec
+			return string.split(/[ \t\n\f\r]+/).filter(function(name) {
+				return name !== ''
+			})
+		}
```
karfau, line 615: I think there is no need to redefine this method every time.
But since the "Infra Spec" [has an own section about this](https://infra.spec.whatwg.org/#split-on-ascii-whitespace), it's probably worth having this peace of code in an extra function on the top level of the module.

Please move it there, but use the following syntax and naming (function name and parameter name), instead of a `var` assignment:
```js
/**
 * @see https://infra.spec.whatwg.org/#split-on-ascii-whitespace
 */
function splitOnASCIIWhitespace (value) {
  // put the implementation here
}
```

lib/dom.js
```diff
@@ -606,14 +606,27 @@ Document.prototype = {
...
 	},
 	
-	getElementsByClassName: function(className) {
-		var pattern = new RegExp("(^|\\s)" + className + "(\\s|$)");
+	getElementsByClassName: function(classNames) {
+		var spaceSeparatedTokens = function(string) {
+			// 5 space characters, as defined in html spec
+			return string.split(/[ \t\n\f\r]+/).filter(function(name) {
+				return name !== ''
+			})
```
karfau, line 614: We are skipping a specific part of the spec in this implementation, that says that the values should not contain duplicates: https://dom.spec.whatwg.org/#concept-ordered-set-parser (it's "hidden" behind the "ordered set" link).

Since we can not rely on `Set` being implemented, the easiest way to avoid duplicates is to use the keys of a dict (object literal).

If you need more details or don't want to bother, I could work on this part after the other changes have been applied. But I would rather add it to this PR, since it basically "adds support for multiple class names".

lib/dom.js
```diff
@@ -606,14 +606,27 @@ Document.prototype = {
...
 			_visitNode(base.documentElement, function(node) {
 				if(node !== base && node.nodeType == ELEMENT_NODE) {
-					if(pattern.test(node.getAttribute('class'))) {
-						ls.push(node);
+					var classActual = node.getAttribute('class')
+					if (classActual && namesTarget.length > 0) { // don't match '' for ''
+						var namesActual = spaceSeparatedTokens(classActual)
+						var matches = namesTarget.every(function(name) {
+							return namesActual.indexOf(name) !== -1
+						})
```
karfau, line 626: I guess that "the most common use case" would be to match a single class name (or even all class names of an element and with some luck in the right order), which could still succeed by just comparing them:
```
// before splitting and iterating just compare them for the most common use case(s)
if (classNames === classActual) {
	ls.push(node);
} else {
	// split and compare as you do above
	if(matches) (
		ls.push(node);
	}
}
```
it's of course an optimization, but I would guess that's it could be worth it performance wise.

karfau:
This PR will also have an impact on #100 since that one tries to add the ability to also call `getElementsByClassName` on any `Element`.

karfau:
@ChALkeR please let us know if you plan to continue to work on this on your own, thx

karfau:
I resolved all the remarks that I added earlier. Thanks again for your contribution.

I needed to force push the rebased branch, other wise github was showing changes to some 30 files and conflicts.

I took the liberty to rename the PR and do a minor tweak to the PR description.

karfau (approved):

PR #215, opened 2021-04-09
docs: Prepare changelog for 0.6.0

karfau:
I created the "Unreleased" section using the `auto-changelog` tool as configured in #180  and tweaked it manually.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,20 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.6.0](https://github.com/xmldom/xmldom/compare/0.5.0...0.6.0)
+
+### Fixes
+
+- Stop serializing empty namespace values like `xmlns:ds=""` [`#168`](https://github.com/xmldom/xmldom/pull/168)  
+  BREAKING CHANGE: If your code expected empty namespaces attributes to be serialized.  
+  Thank you [@pdecat](https://github.com/pdecat) and [@FranckDepoortere](https://github.com/FranckDepoortere)
+- Escape `<` to `&lt;` when serializing attribute values [`#198`](https://github.com/xmldom/xmldom/issues/198) / [`#199`](https://github.com/xmldom/xmldom/pull/199)
+
 ## 0.5.0
 
 ### Fixes
 - Avoid misinterpretation of malicious XML input - [`GHSA-h6q6-9hqw-rwfv`](https://github.com/xmldom/xmldom/security/advisories/GHSA-h6q6-9hqw-rwfv) (CVE-2021-21366)
-  - Improve error reporting; throw on duplicate attribute
+  - Improve error reporting; throw on duplicate attribute\
     BREAKING CHANGE: It is currently not clear how to consistently deal with duplicate attributes, so it's also safer for our users to fail when detecting them.
     It's possible to configure the `DOMParser.errorHandler` before parsing, to handle those errors differently.
 


CHANGELOG.md
```diff
@@ -4,11 +4,20 @@ All notable changes to this project will be documented in this file.
...
+  BREAKING CHANGE: If your code expected empty namespaces attributes to be serialized.  
+  Thank you [@pdecat](https://github.com/pdecat) and [@FranckDepoortere](https://github.com/FranckDepoortere)
+- Escape `<` to `&lt;` when serializing attribute values [`#198`](https://github.com/xmldom/xmldom/issues/198) / [`#199`](https://github.com/xmldom/xmldom/pull/199)
+
 ## 0.5.0
 
 ### Fixes
 - Avoid misinterpretation of malicious XML input - [`GHSA-h6q6-9hqw-rwfv`](https://github.com/xmldom/xmldom/security/advisories/GHSA-h6q6-9hqw-rwfv) (CVE-2021-21366)
-  - Improve error reporting; throw on duplicate attribute
+  - Improve error reporting; throw on duplicate attribute\
```
brodybits, line 20: ??

CHANGELOG.md
```diff
@@ -4,11 +4,20 @@ All notable changes to this project will be documented in this file.
...
 
+## [0.6.0](https://github.com/xmldom/xmldom/compare/0.5.0...0.6.0)
+
+### Fixes
+
+- Stop serializing empty namespace values like `xmlns:ds=""` [`#168`](https://github.com/xmldom/xmldom/pull/168)  
+  BREAKING CHANGE: If your code expected empty namespaces attributes to be serialized.  
+  Thank you [@pdecat](https://github.com/pdecat) and [@FranckDepoortere](https://github.com/FranckDepoortere)
+- Escape `<` to `&lt;` when serializing attribute values [`#198`](https://github.com/xmldom/xmldom/issues/198) / [`#199`](https://github.com/xmldom/xmldom/pull/199)
+
```
brodybits, line 15: LGTM

CHANGELOG.md
karfau: That's adding a missing line break.
I can also replace it with two spaces if you prefer

karfau:
@brodybits Since I never released a version of this lib before and I'm not aware of any documentation, of how you prefer it happens, can you do it or do you want me to give it a try?

karfau:
@brodybits can we land this and get the release out there?

CHANGELOG.md
brodybits: Let's keep the backslash (`\`) then. I did check that adding the backslash *does* add the line break.

brodybits:
@karfau I totally missed your question last week, my apologies. I generally update the changelog in the same commit as I make the release, see these for example:

- 9df224b212f8c3fab60d0a688679b16166a6a862 - version `0.4.0`
- f763b002fb01d3e58010b04be35a3be50f9777e8 - version `0.5.0`

Then I would of course create the release with the info from the changelog according to https://docs.github.com/en/github/administering-a-repository/managing-releases-in-a-repository#creating-a-release

and then do `npm publish` to publish on npm.

I can publish this release next week if needed but would prefer it if you can get on board with this task. I would be happy to answer any questions here, privately, or set up a Zoom session if needed.

karfau:
So if I understood you correctly this would be the commands to execute for this release:

```bash
git checkout master # pointing to github.com/karfau/xmldom, update to xmldom/xmldom@master if required
git checkout 0.6.0/changelog -- CHANGELOG.md
git add CHANGELOG.md
npm version minor --force -m "xmldom version 0.6.0" # should commit and create git tag
git push --tags upstream master # upstream is xmldom/xmldom in my case
# create a release on github for the newly added tag
npm publish
```

(According to [the docs](https://docs.npmjs.com/cli/v7/commands/npm-version), the `--force` for `npm version` seems to be required since with the updated changelog the working directory is not clear.)

karfau:
It didn't went so smooth:

On the first attempt my global pre-commit hook that prevents me from commiting to master by accident of course prevented the commit from `npm version`, after disabling it and running again I locally had a version `0.7.0`. After revert and retry it worked, but the tag was `v0.6.0` so I needed to create the one for `0.6.0` manually. (I dropped the `v0.6.0` one that I had already pushed.)
But I think/hope that now everything is correct and I was able to publish the new version :tada: 

Closing this PR since it's no longer required. But I really think we should make this process more easy/less manual.

karfau:
OK, `git push` didn't work as I expected it to work, so the tag I created was on the wrong commit, I corrected it now.

brodybits:
@karfau thanks for making the release.

I think you are absolutely correct that we should look for a way to improve this process. I would recommend that we take a look at how some other projects have improved the automation.

PR #229, opened 2021-04-25
fix: Set `localName` as part of `Document.createElement`

rrthomas:
This seemed to be missing. Without it, xpath (XPath 1.0) works, but
xpath2 (XPath 2.0) and fontoxpath (XPath 3.x) do not match elements.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -627,6 +627,7 @@ Document.prototype = {
 		node.ownerDocument = this;
 		node.nodeName = tagName;
 		node.tagName = tagName;
+		node.localName = tagName;
 		node.childNodes = new NodeList();
 		var attrs	= node.attributes = new NamedNodeMap();
 		attrs._ownerElement = node;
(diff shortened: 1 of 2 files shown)


karfau:
I just checked, the property `Element.localName` (and also setting that value as part of `Document.createElement`) is only specified in the DOM living standard, but [`Node.localName` is already specified in DOM Level 2](https://www.w3.org/TR/DOM-Level-2-Core/#core-ID-NodeNSLocalN).

I also checked that we are already and in all cases setting this field as part of `Document.createElementNS` (so using that method instead could be a workaround, until this PR is merged and released.

@rrthomas Thank you for the PR, are you able to add some minimal test(s) to `test/dom/element.test.js` that make sure this doesn't break again? (You can add a nested describe with a message of `"createElement"` and add your tests to it.)

rrthomas:
Thanks for the analysis (I'm not an expert in this area, so didn't venture to go further on my own!). I'll see if I can add a test.

karfau:
Thank you for your contribution, it will be part of the next release.

PR #233, opened 2021-05-09
refactor: Improve exports & require statements

brodybits:
by creating a proper main entry file `lib/index.js`.
NOTE that the following exports have been deprecated in `lib/dom-parser.js` and will be removed in the next minor release:
- `DOMImplementation`
- `XMLSerializer`

### DIFF
--- a/lib/dom-parser.js
+++ b/lib/dom-parser.js
@@ -254,7 +254,10 @@ function appendElement (hander,node) {
 }//appendChild and setAttributeNS are preformance key
 
 //if(typeof require == 'function'){
-exports.DOMImplementation = require('./dom').DOMImplementation;
+/**
+ * @deprecated Import/require from main entry point instead
+ */
+exports.DOMImplementation = dom.DOMImplementation;
 exports.XMLSerializer = require('./dom').XMLSerializer ;
 exports.DOMParser = DOMParser;
 exports.__DOMHandler = DOMHandler;
--- a/lib/dom-parser.js
+++ b/lib/dom-parser.js
@@ -1,7 +1,13 @@
 var conventions = require("./conventions");
 var entities = require('./entities');
 
+var DOMImplementation = require('./dom').DOMImplementation;
+var sax = require('./sax');
+
 var NAMESPACE = conventions.NAMESPACE;
+
+var XMLReader = sax.XMLReader;
+var ParseError = sax.ParseError;
 
 function DOMParser(options){
 	this.options = options ||{locator:{}};
@@ -247,10 +253,6 @@ function appendElement (hander,node) {
 }//appendChild and setAttributeNS are preformance key
 
 //if(typeof require == 'function'){
-var sax = require('./sax');
-var XMLReader = sax.XMLReader;
-var ParseError = sax.ParseError;
-var DOMImplementation = require('./dom').DOMImplementation;
 exports.DOMParser = DOMParser;
 exports.__DOMHandler = DOMHandler;
 //}
--- a/lib/dom-parser.js
+++ b/lib/dom-parser.js
@@ -252,5 +252,7 @@ function appendElement (hander,node) {
     }
 }//appendChild and setAttributeNS are preformance key
 
+//if(typeof require == 'function'){
 exports.DOMParser = DOMParser;
 exports.__DOMHandler = DOMHandler;
+//}
--- a/lib/dom-parser.js
+++ b/lib/dom-parser.js
@@ -7,8 +7,8 @@ var DOMImplementation = dom.DOMImplementation;
 
 var NAMESPACE = conventions.NAMESPACE;
 
-var XMLReader = sax.XMLReader;
 var ParseError = sax.ParseError;
+var XMLReader = sax.XMLReader;
 
 function DOMParser(options){
 	this.options = options ||{locator:{}};
--- a/lib/dom-parser.js
+++ b/lib/dom-parser.js
@@ -253,15 +253,15 @@ function appendElement (hander,node) {
     }
 }//appendChild and setAttributeNS are preformance key
 
-//if(typeof require == 'function'){
+exports.__DOMHandler = DOMHandler;
+exports.DOMParser = DOMParser;
+
 /**
  * @deprecated Import/require from main entry point instead
  */
 exports.DOMImplementation = dom.DOMImplementation;
+
 /**
  * @deprecated Import/require from main entry point instead
  */
 exports.XMLSerializer = dom.XMLSerializer;
-exports.DOMParser = DOMParser;
-exports.__DOMHandler = DOMHandler;
-//}
--- a/lib/dom-parser.js
+++ b/lib/dom-parser.js
@@ -250,8 +250,7 @@ function appendElement (hander,node) {
 var sax = require('./sax');
 var XMLReader = sax.XMLReader;
 var ParseError = sax.ParseError;
-var DOMImplementation = exports.DOMImplementation = require('./dom').DOMImplementation;
-exports.XMLSerializer = require('./dom').XMLSerializer ;
+var DOMImplementation = require('./dom').DOMImplementation;
 exports.DOMParser = DOMParser;
 exports.__DOMHandler = DOMHandler;
 //}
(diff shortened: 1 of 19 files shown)


karfau (commented):
My understanding of what this change is good for is very limited. Maybe you can elaborate on the problem this solves in the PR description or add a link to something that explains it?

From what I know, such a file can be used to declare the "public API" from the "internal" exports of other modules.
But if that is the goal, I don't really understand why (all) the tests should go through  the public API.
We also never discussed what thing should be considered "public" and which not. Which is relevant for me when we need to resolve conflicts with the PRs that add exports to some modules in lib.

karfau:
> BREAKING CHANGE due to exports removed from `lib/dom-parser.js`

Just thinking out loud:
Since lib/dom-parser.js was the main entry point when importing/requiring `xmldom` can we assume that nobody would write `xmldom/lib/dom-parser` instead?
If that is the case, since all the exports that previously existed an the main entry point also exist on the new one, this is not a breaking change? (Or did I miss something?)
I also think when doing breaking changes we should be more explicit by listing the exports that have been removed.

karfau:
@brodybits 
&TLDR; I would love to land this asap, but as long as it's a breaking change, it's not that easy.

1. Can you please reply to [my question](https://github.com/xmldom/xmldom/pull/233#issuecomment-837784090)?
2. could you either rebase this branch and resolve the conflicts or push the branch to the xmldom repo, so it's easier for me to continue working on it?

Thx

brodybits:
> Since lib/dom-parser.js was the main entry point when importing/requiring `xmldom` can we assume that nobody would write `xmldom/lib/dom-parser` instead?

*probably* yes

> I also think when doing breaking changes we should be more explicit by listing the exports that have been removed.

Done in the now updated description (it was just `XMLSerializer`).

Yeah I don't think this would likely break any real users, definitely should not be in a patch release though.

> 2\. could you either rebase this branch and resolve the conflicts or [...]

I will resolve merge and/or rebase to resolve the conflicts.

In terms of coding it would probably be easier to start this over. But merge & rebase is better for the PR flow. Coming up.

karfau:
The exports that you listed have been removed from `lib/dom-parser.js` which was the main enrty point, but that are still present on the main entry point, right?

I would argue that importing things from `@xmldom/xmldom/lib/dom-parser` (instead of just `@xmldom/xmldom`) should not be considered part of the "public API", so I wouldn't consider it a breaking change.

And if we can not agree on that, we could make it non breaking **for now** by restoring those exports and marking them as `@deprecated`, so people can become aware and we can land it and release it as a patch version, and only remove them in a future PR.

What do you think?

lib/dom-parser.js
```diff
@@ -1,8 +1,14 @@
 var conventions = require("./conventions");
 var entities = require('./entities');
 
+var DOMImplementation = require('./dom').DOMImplementation;
+var sax = require('./sax');
```
karfau, line 5: nit: I would love to see all require statements in one block and ordered alphabetically.

```suggestion
var dom = require('./dom')
var entities = require('./entities');
var sax = require('./sax');

var DOMImplementation = dom.DOMImplementation;
```

lib/dom-parser.js
```diff
@@ -1,8 +1,14 @@
...
+
 var NAMESPACE = conventions.NAMESPACE;
 
+var XMLReader = sax.XMLReader;
+var ParseError = sax.ParseError;
```
karfau, line 10: nit: Again just alphabetical order of "imports"
```suggestion
var ParseError = sax.ParseError;
var XMLReader = sax.XMLReader;
```

lib/dom-parser.js
```diff
@@ -247,10 +253,7 @@ function appendElement (hander,node) {
 }//appendChild and setAttributeNS are preformance key
 
 //if(typeof require == 'function'){
-var sax = require('./sax');
-var XMLReader = sax.XMLReader;
-var ParseError = sax.ParseError;
-var DOMImplementation = exports.DOMImplementation = require('./dom').DOMImplementation;
+exports.DOMImplementation = require('./dom').DOMImplementation;
```
karfau, line 256: nit: in alignment with the previous suggestions:
```suggestion
/**
 * @deprecated Import/require from main entry point instead
 */
exports.DOMImplementation = dom.DOMImplementation;
```

lib/dom-parser.js
```diff
@@ -247,10 +253,7 @@ function appendElement (hander,node) {
...
 exports.XMLSerializer = require('./dom').XMLSerializer ;
```
karfau, line 257: nit: in alignment with the previous suggestions:

```suggestion
/**
 * @deprecated Import/require from main entry point instead
 */
exports.XMLSerializer = dom.XMLSerializer ;
```

lib/index.js
```diff
@@ -0,0 +1,3 @@
+exports.DOMImplementation = require('./dom').DOMImplementation
+exports.XMLSerializer = require('./dom').XMLSerializer
```
karfau, line 2: Question: does it make sense/ a difference to have a single vs multiple `require('./dom')` in a single file?
```suggestion
var dom = require('./dom')
exports.DOMImplementation = dom.DOMImplementation
exports.XMLSerializer = dom.XMLSerializer
```

test/dom/serializer.test.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-const { DOMParser } = require('../../lib/dom-parser')
+const { DOMParser } = require('../../lib')
 const { XMLSerializer } = require('../../lib/dom')
```
karfau, line 4: nit: They could be merged now
```suggestion
const { DOMParser, XMLSerializer } = require('../../lib')
```

karfau (approved):
I would love to see my suggestions applied, but I would not enforce it.

lib/dom-parser.js
```diff
@@ -247,11 +254,14 @@ function appendElement (hander,node) {
...
+/**
+ * @deprecated Import/require from main entry point instead
+ */
+exports.DOMImplementation = dom.DOMImplementation;
+/**
+ * @deprecated Import/require from main entry point instead
+ */
+exports.XMLSerializer = dom.XMLSerializer;
 exports.DOMParser = DOMParser;
 exports.__DOMHandler = DOMHandler;
```
brodybits, line 266: @karfau this PR made me start to wonder why we are exporting DOMHandler with `__` prefix. I recall there was some reason for this when we made the security update for `0.5.0` but embarrassed I do not recall exactly. Can you help me understand this again?

karfau:
I assume you will mark it as ready for review and land it when you are ready :)

lib/dom-parser.js
karfau: Sure, it was my primitive way to indicate a "private API" and to be able to use it in tests.

Now that we have a clear separation of public (`lib/index.js`) and private (all other modules) API, we could drop the `__` and just export/require `DOMHandler`.

brodybits: Oh yeah !!

I am thinking that if we drop the `__` here, it could be easy for a reader to confuse between what is exported internally (for internal testing) and what is to be exported externally.

I am also thinking it would be ideal if we could find a way to make the pluggable DOMHandler usable from the external API somehow.

There are some (myself included) who think that test mocking is a kind of code smell: https://medium.com/javascript-scene/mocking-is-a-code-smell-944a70c90a6a

I would like to propose that we just leave it for now and revisit someday later.

brodybits:
@karfau can you take one last look at the exports before I merge this?

karfau (approved):
Perfect, let's merge this!

lib/dom-parser.js
karfau: I'm fine with that.
I think mocking needs to be applied with much care, which I hope I did, to avoid brittle and unmaintainable tests.
But of course I'm open to suggestions on how to test those things in a different way.

brodybits:
@karfau I accidentally removed you from co-author credit when I merged this, my apologies. I don't know if there is anything we can do about this in git. Argh!!

karfau:
@brodybits :laughing: no problem, you did the main work and I have more than enough credit(s?) in this project

PR #247, opened 2021-06-06
docs(readme): Link to the raw SVG directly

karfau:
currently when clicking on the image the github page showing the rich preview of the SVG opens, which still doesn't allow to click on the contained links.

By linking to the raw file on github, clicking on the links is possible after the first click.
Of course the link will break if it is renamed and always point's to the master branch, no matter which branch you are on...
I was not aware, that links in SVGs work in this limited way on github.


brodybits (commented):
It would be especially cool if each link in the SVG would open in a new tab. (If I click on the SVG link in my preview of README.md, it *does* seem to open the SVG itself in a new tab.)

karfau:
@brodybits I tried it with `target="_blank"` but in chrome those link are not opened at all because of some security restrictions that I was not able to configure in a way that the links would work.
So I decided to revert.

PR #278, opened 2021-08-19
chore(npm): Scope package to @xmldom org

karfau:
As discussed in the recent days, this PR prepares the scoping of the package into the npm organisation `xmldom` so the new package name is `@xmldom/xmldom`.

See https://docs.npmjs.com/about-scopes
I also checked that the `@` has to be part of the shields/badges URL to work, by looking at https://shields.io/category/version and trying it with an existing scoped package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,11 +57,6 @@
       "name": "brodybits",
       "email": "chris@brody.consulting",
       "url": "https://github.com/brodybits"
-    },
-    {
-      "name": "Christian Bewernitz",
-      "email": "coder@karfau.de",
-      "web": "https://github.com/karfau"
     }
   ],
   "contributors": [
@@ -89,6 +84,11 @@
       "name": "Eric Newport",
       "email": "kethinov@gmail.com",
       "web": "https://github.com/kethinov"
+    },
+    {
+      "name": "Christian Bewernitz",
+      "email": "coder@karfau.de",
+      "web": "https://github.com/karfau"
     }
   ],
   "bugs": {
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "xmldom",
+  "name": "@xmldom/xmldom",
   "version": "0.7.0",
   "description": "A pure JavaScript W3C standard-based (XML DOM Level 2 Core) DOMParser and XMLSerializer module.",
   "keywords": [
@@ -57,6 +57,11 @@
       "name": "brodybits",
       "email": "chris@brody.consulting",
       "url": "https://github.com/brodybits"
+    },
+    {
+      "name": "Christian Bewernitz",
+      "email": "coder@karfau.de",
+      "web": "https://github.com/karfau"
     }
   ],
   "contributors": [
@@ -84,11 +89,6 @@
       "name": "Eric Newport",
       "email": "kethinov@gmail.com",
       "web": "https://github.com/kethinov"
-    },
-    {
-      "name": "Christian Bewernitz",
-      "email": "coder@karfau.de",
-      "web": "https://github.com/karfau"
     }
   ],
   "bugs": {
--- a/readme.md
+++ b/readme.md
@@ -1,10 +1,14 @@
-# XMLDOM
+# @xmldom/xmldom
 
-[![license](https://img.shields.io/npm/l/xmldom?color=blue&style=flat-square)](LICENSE)
-[![npm](https://img.shields.io/npm/v/xmldom?style=flat-square)](https://www.npmjs.com/package/xmldom)
+[![license](https://img.shields.io/npm/l/@xmldom/xmldom?color=blue&style=flat-square)](LICENSE)
+[![npm](https://img.shields.io/npm/v/@xmldom/xmldom?style=flat-square)](https://www.npmjs.com/package/@xmldom/xmldom)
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -57,6 +57,11 @@
       "name": "brodybits",
       "email": "chris@brody.consulting",
       "url": "https://github.com/brodybits"
+    },
+    {
+      "name": "Christian Bewernitz",
+      "email": "coder@karfau.de",
+      "web": "https://github.com/karfau"
```
brodybits, line 64: nit: This update seems orthogonal (unrelated) to me from actually changing the scope to use the @xmldom org.

I start to wonder if we should remove the maintainers & contributors entries when we make the next minor release (0.8.0). I have a feeling that this kind of info should be available from npm & GitHub, would be nice if we don't have to keep synchronization as maintainers come and go in the future.

readme.md
```diff
@@ -1,11 +1,15 @@
-# XMLDOM
+# @xmldom/xmldom
```
brodybits, line 1: 👍 thanks!

package.json
karfau: You are right, thx for pointing it out.
Reverted the related changes.
I like the idea of not synchronizing thisinformation into `package.json`.

brodybits:
I am starting to have some "dumb" questions:

I am assuming that this to be a squash merge commit.

The first version under the `@xmldom/xmldom` scope would be `0.7.0` ... right?

Contrary to what I have seen discussed elsewhere, I would be a bit reluctant to publish updates for anything pre-0.7.0. I would certainly consider the introduction of package scope to be a breaking change by itself; previous security updates needed breaking changes; impact of breaking changes up to `0.7.0` seem to be pretty limited.

Which brings up another question: if we already have a `0.7.0` tag without the @xmldom org scope, and people may have already starting using the `0.7.0` tag from GitHub, then what tag should we use on GitHub once we change the scope to `@xmldom/xmldom`?

Shouldn't we consider bumping the version to `0.8.0` to help avoid potential confusion?

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "xmldom",
+  "name": "@xmldom/xmldom",
```
brodybits, line 2: 👍 now agreed on my part, should maximize consistency with the GitHub project URL and how this project would normally be installed from GitHub: `npm:xmldom/xmldom` or `npm:xmldom/xmldom#version` (`version` could be branch, tag, or commit ID)

karfau:
I will do a bit of research how other packages are handling this. Since the only change is the package name i would rather consider it a patch version change. I'm also on favor of not updaying the existing 0.7.0 tag.

It's true that the argument of unattended update due to version ranges in package.json would not work.

But providing the patched versions (which I already prepared), make it easier to just switch to the scoped package, without the need to consider our recent breaking changes.

joebowbeer (approved):
LGTM

brodybits:
While I would agree with the notion that this update is very limited in scope, I would still consider it to be a potentially breaking change. If someone would do an install from GitHub, the package name under dependencies would likely change as well ... and the import statement would need to be updated in the downstream source code.

karfau:
But I think we could also just (git) tag it as `@xmldom/xmldom@0.7.0`

brodybits:
And what about the next release?

I would downvote the scoped tagging in general. This would add some ugliness to releases from GitHub, installing by tag from GitHub, downloading ZIP of a release from GitHub, etc. etc.

karfau:
Changing all the usages in source code could  be avoided by putting the following into package.json:
`"xmldom":"@xmldom/xmldom@0.7.0"` (the version doesn't really matter here)

I think I'm still more convinced by the idea that the same code being  published in different places can happily have the same versions.

brodybits (approved):
Good to go, thanks @karfau.

I would continue to downvote changing the package name in the patch release, not 100% sure if all recent npm versions will work properly with this. I would favor putting the scoped package name together with PR #279 (removing extra author / maintainer info) in a new `0.8.0` release. But I think my sentiment should not block merging and publishing these updates.

forty:
I think renaming the package makes all versioning considerations irrelevant. LGTM, thanks!

karfau:
@brodybits a version tag that would be semver compatible would be to add something as the [build metadata](https://semver.org/#spec-item-10). Since `@` and `/` are not allowed we could just add `0.7.0+unscoped` to the same commit that is now `0.7.0` and `0.7.0+scoped` to the commit from this PR and document the meaning in the changelog.
In the future I would just use version numbers without any additional metadata, since there is only the scoped version being released.

brodybits:
Thanks @karfau. I had thought that npm did not work so well with the suffix starting with "+" ... but if it works, great!

I think I understand your proposal after multiple read-throughs. So here is how I think this would play out for a while starting with `0.6.0`, leaving patch releases out of consideration:

- existing tag `0.6.0` - unscoped `xmldom` package version `0.6.0`, already published to npm
- existing tag `0.7.0` - unscoped `xmldom` package version `0.7.0` ... not published to npm, no plan to publish to npm
- upcoming new tag `0.7.0+unscoped` - same as existing tag `0.7.0` above
- upcoming new tag `0.7.0+scoped` - plan to publish to npm as `@xmldom/xmldom` version `0.7.0`
- future version `0.8.0` with tag `0.8.0` - would be published to npm as `@xmldom/xmldom` version `0.8.0`

I think this means that upgrades from npm will need special handling for a while, and anyone installing from GitHub may also need to do some special handling for a while as well.

No objections on my part, just wanted to think it through a little before we move forward.

karfau:
Yes, @brodybits this is exactly what I had in mind, thx for elaborating it this precisely.

I will link to this comment in the changelog :wink:

PR #279, opened 2021-08-19
chore(package.json): remove author, maintainers, etc.

brodybits:
since they are visible on github/npm level and don't need to additionally be synced/maintained inside `package.json`, as discussed for followup in PR #278.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,6 @@
     "XMLSerializer",
     "ponyfill"
   ],
-  "author": "jindw <jindw@xidea.org> (http://www.xidea.org)",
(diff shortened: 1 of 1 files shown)


karfau (commented):
@brodybits I wonder if it would make sense to also drop the "author" property.
At least it feels strange that it points to somebody that is no longer involved/we have not been able to contact and it being the only direct contact information left in this file.

brodybits:
Agreed on my part to remove the authors as well, thanks.

karfau (approved):

PR #280, opened 2021-08-19
Revert "chore(deps): update @stryker-mutator/core -> ^5.3.0 - devdependencies"

karfau:
Reverts xmldom/xmldom#273

since the checks fail on the master branch.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -121,96 +121,17 @@
       }
     },
     "@babel/helper-create-class-features-plugin": {
-      "version": "7.15.0",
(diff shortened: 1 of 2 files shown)


brodybits:
Shouldn't we analyze *why* this in-range update suddenly failed ... and ideally if there is anything we can do to avoid this kind of thing in the future (root-cause analysis)?

Or am I just being over-analytical?

karfau:
Yes, we should, but we can do that locally without having a red master build.
And it's not on the top of my prio list right now.

(I usually use the [dependency dashboard](https://docs.renovatebot.com/configuration-options/#dependencydashboard) feature with renovate, I just realized that this is not in place in our case. Will add it later, so it's easier to keep track of those reverts/closed update PRs.)

PR #291, opened 2021-08-21
fix: Normalize white space characters and preserve references in attributes

### LINKED ISSUE #284: Incorrect serialization of XML attributes containing white space
The code for serializing attributes (here: <https://github.com/xmldom/xmldom/blob/6ce470094c7ea78baa4e5ea18e912a88cd11a63c/lib/dom.js#L1138-L1140>) does not treat newlines in any special way.

This is a problem, because XML parses literal newlines as spaces. However, it allows newlines in attributes when they are provided as character references. (See: <https://www.w3.org/TR/xml/#AVNormalize>.)

So:

```xml
<element attribute="with&#xA;newline"></element>
```

serializes as

```xml
<element attribute="with
newline"></element>
```

which then gets parsed as

```xml
<element attribute="with newline"></element>
```

which is not equivalent.

karfau:
- Added normalization of white space literal characters as specified in https://www.w3.org/TR/xml/#AVNormalize, they now appear as space(s) in the attribute values when accessed from the DOM (and are also serialized as space)
- Added serialization of white space literal values to to the correct numerical character references as described in https://github.com/w3c/DOM-Parsing/issues/59 to fix #284 

BREAKING CHANGE: If you relied on the not spec compliant preservation of `\t`, `\n` or `\r` in **attribute values**. To preserve those you will have to create XML that instead contains the correct numerical (or hexadecimal) equivalent (`&#x9;`, `&#xA;`, `&#xD;`).

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -1142,7 +1142,7 @@ function needNamespaceDefine(node, isHTML, visibleNamespaces) {
  * @see https://www.w3.org/TR/xml/#NT-AttValue
  */
 function addSerializedAttribute(buf, qualifiedName, value) {
-	buf.push(' ', qualifiedName, '="', value.replace(/[<&"]/g,_xmlEncoder), '"')
+	buf.push(' ', qualifiedName, '="', value.replace(/[<&"\t\n\r]/g,_xmlEncoder), '"')
 }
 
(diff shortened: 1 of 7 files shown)


karfau:
Such a mess... the subsequent PR landing seems to only work when the branches are all in the same repository.
So I rebased and recreated the PR in #310

PR #297, opened 2021-08-23
chore: Split test and lint scripts

### LINKED ISSUE #111: keep lint in separate GitHub action
idea from @karfau

Prettier does this, and I think it would help people see more quickly whether a build failure comes from lint or from a test failure.

karfau:
To fix #111 I think it's good enough to split test and lint into two different build steps.
Separate actions are not required since the github actions details view now shows what steps failed.
image

To be able to do that split, I needed to drop the execution of the `lint` script from the `test` script. The `start` script still runs both in watch mode, so it's still easy to run both when developing.

### DIFF
--- a/.github/workflows/test-node.js.yml
+++ b/.github/workflows/test-node.js.yml
@@ -28,4 +28,6 @@ jobs:
       uses: actions/setup-node@v2
       with:
         node-version: ${{ matrix.node-version }}
-    - run: npm cit
+    - run: npm ci --no-audit
+    - run: npm run test
+    - run: npm run lint
--- a/package-lock.json
+++ b/package-lock.json
@@ -1922,15 +1922,6 @@
       "integrity": "sha512-0ISdNousHvZT2EiFlZeZAHBUvSxmKswVCEf8hW7KWgG4a8MVEu/3Vb6uWYozkjylyCxe0JBIiRB1jV45S70WVQ==",
       "dev": true
     },
-    "define-properties": {
-      "version": "1.1.3",
-      "resolved": "https://registry.npmjs.org/define-properties/-/define-properties-1.1.3.tgz",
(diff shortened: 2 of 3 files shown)


brodybits (commented):
It looks to me like npm-run-all is no longer needed and should be removed ?

karfau:
@brodybits you are right, I dropped it, quite some reduction on devDependencies side
```
$ npm rm npm-run-all
removed 49 packages, updated 5 packages and audited 600 packages in 12.223s
```

.github/workflows/test-node.js.yml
```diff
@@ -28,4 +28,6 @@ jobs:
       uses: actions/setup-node@v2
       with:
         node-version: ${{ matrix.node-version }}
-    - run: npm cit
+    - run: npm ci --no-audit
```
brodybits, line 31: minor question: Can you quickly describe the motivation behind adding `--no-audit` here?

brodybits (approved):
LGTM overall

.github/workflows/test-node.js.yml
karfau: Yes:
a) (Since we have tools to provide update PRs) nobody is looking at the audit output on CI
b) [it saves time to not do it](http://www.tiernok.com/posts/2019/faster-npm-installs-during-ci/)

brodybits:
@karfau it looks like this broke the Stryker run in master ... definitely not the first time we hit this kind of thing.

I am wondering if using Stryker Jest runner, as I tried in PR #208, could help avoid this failure.

I am thinking now if we could find a way to add another task to check that Stryker can do its initial run in the future PRs?

karfau:
> I am thinking now if we could find a way to add another task to check that Stryker can do its initial run in the future PRs?

Yes, also thought about this, but I couldn't find any documentation about such a "dryrun" mode yet.

PR #302, opened 2021-08-25
ci: Execute stryker:dryrun on branches

karfau:
Dryrun works by passing a glob that matches no file.
Additionally we reduce the reporters since there is nothing to report anyways.
I decided to go for a CLI approach instead of a separate config file,
since there is no way to extend config files, and I would like to avoid that we need to keep the two files in sync.

https://github.com/stryker-mutator/stryker-js/issues/3088


karfau:
@brodybits I'm thinking about actually splitting the github actions files further, since I think it's actually not required to run `npm run lint` on every tested node version.

PR #304, opened 2021-08-26
ci(github): Split test and lint steps into jobs

### LINKED ISSUE #111: keep lint in separate GitHub action

karfau:
to allow linting to be listed as a separately failing check in the PR.

Really fixes #111

Note: I don't think there is any benefit of splitting them into different files as of now.

### DIFF
--- a/.github/workflows/test-node.js.yml
+++ b/.github/workflows/test-node.js.yml
@@ -20,7 +20,7 @@ jobs:
         - 10.x
         - 12.x
         - 14.x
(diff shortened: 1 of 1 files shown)


karfau:
@brodybits Do you think we can land this?

brodybits:
I may need another 1-2 days to look at these, my apologies.

Issue #303, opened 2021-08-26
xmldom is not normalizing line endings before parsing


karfau:
Specs: https://www.w3.org/TR/xml/#sec-line-ends

> 2.11 End-of-Line Handling
> XML parsed entities are often stored in computer files which, for editing convenience, are organized into lines.  These lines are typically separated by some combination of the characters CARRIAGE RETURN (#xD) and LINE FEED (#xA).
> 
> To simplify the tasks of applications, the XML processor must behave as if it normalized all line breaks in external parsed entities (including the document entity) on input, before parsing, by translating both the two-character sequence #xD #xA and any #xD that is not followed by #xA to a single #xA character

PR #305, opened 2021-08-26
fix: Normalize line breaks according to spec

karfau:
> XML parsed entities are often stored in computer files which, for editing convenience, are organized into lines. These lines are typically separated by some combination of the characters CARRIAGE RETURN (#xD) and LINE FEED (#xA).
> 
> To simplify the tasks of applications, the XML processor must behave as if it normalized all line breaks in external parsed entities (including the document entity) on input, before parsing, by translating both the two-character sequence #xD #xA and any #xD that is not followed by #xA to a single #xA character.

Where `#xD` == `\r` and `#xA` == `\n`, so 
` \r\n ` => ` \n `
` \n\r ` => ` \n\n `
` \n  ` => ` \n `
` \r  ` => ` \n `

BREAKING CHANGE: Certain combination of line break characters are normalized before parsing takes place and will not longer be preserved. For details see https://www.w3.org/TR/xml/#sec-line-ends

fixes https://github.com/xmldom/xmldom/issues/303

### DIFF
--- a/lib/dom-parser.js
+++ b/lib/dom-parser.js
@@ -33,10 +33,14 @@ DOMParser.prototype.parseFromString = function(source,mimeType){
 		defaultNSMap[''] = NAMESPACE.HTML;
 	}
 	defaultNSMap.xml = defaultNSMap.xml || NAMESPACE.XML;
-	if(source && typeof source === 'string'){
-		sax.parse(source,defaultNSMap,entityMap);
-	}else{
-		sax.errorHandler.error("invalid doc source");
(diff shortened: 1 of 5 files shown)


karfau:
Thank you again for the review, I had a "translation error" in my first attempt: I mixed up `\r` and `\n`...

Sorry for the back end forth, I will recreate this PR, since I want #291 to be based on this change I need to push these changes to this repo not my fork. (This way when landing this change the other PR will automatically be switched to be based on master.)

PR #307, opened 2021-08-28
fix: Normalize line breaks according to spec

karfau:
> XML parsed entities are often stored in computer files which, for editing convenience, are organized into lines. These lines are typically separated by some combination of the characters CARRIAGE RETURN (#xD) and LINE FEED (#xA).
>
> To simplify the tasks of applications, the XML processor must behave as if it normalized all line breaks in external parsed entities (including the document entity) on input, before parsing, by translating both the two-character sequence #xD #xA and any #xD that is not followed by #xA to a single #xA character.

Where `#xD` == `\r` and `#xA` == `\n`, so
` \r\n ` => ` \n `
` \n\r ` => ` \n\n `
` \n  ` => ` \n `
` \r  ` => ` \n `

BREAKING CHANGE: Certain combination of line break characters are normalized before parsing takes place and will no longer be preserved. For details see https://www.w3.org/TR/xml/#sec-line-ends

fixes https://github.com/xmldom/xmldom/issues/303

### DIFF
--- a/test/parse/test-doc-whitespace.test.js
+++ b/test/parse/test-doc-whitespace.test.js
@@ -20,7 +20,7 @@ describe('errorHandle', () => {
 		expect({ actual, ...errors }).toMatchSnapshot()
 	})
 
-	it('test', () => {
+	it('should encode < literal when not part of a tag', () => {
 		const description = '<p>populaciji (< 0.1%), te se</p>'
 		const { errors, parser } = getTestParser()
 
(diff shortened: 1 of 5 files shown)


karfau:
@brodybits I would like to land this one next, to avoid rebasing #291 over and over.

test/parse/test-doc-whitespace.test.js
```diff
@@ -29,3 +29,57 @@ describe('errorHandle', () => {
...
+			'#xa#x20#xa',
+		],
+		[
+			'in firstChild text node',
+			'<xml> \r\r<child/></xml>',
+			(dom) => dom.documentElement.firstChild.nodeValue,
+			'#x20#xa#xa',
+		],
+	])(
+		'should normalize "\\r" not followed by "\\n" %s',
```
brodybits, line 76: nit: should this be tested in attribute as well?

test/parse/test-doc-whitespace.test.js
```diff
@@ -29,3 +29,57 @@ describe('errorHandle', () => {
 		expect({ actual, ...errors }).toMatchSnapshot()
 	})
 })
+
+describe('whitespace', () => {
+	const whitespaceToHex = (str) =>
```
brodybits, line 34: question: could this be a more general-purpose reusable function?

Or am I violating this by suggesting a too-hasty abstraction: https://kentcdodds.com/blog/aha-programming

brodybits (approved):
LGTM thanks!

I think it would be nice to take care of the minor nit, don't think it needs to be blocking though.

test/parse/test-doc-whitespace.test.js
karfau: It can not easily, since the the algorithm in the other PR would replace it with a space.
But since we are replacing all occurrences in one go before parsing, I don't think it matters to much.

test/parse/test-doc-whitespace.test.js
karfau: I was also thinking about it. I think as soon as it is needed elsewhere we will find a proper place for it.

PR #310, opened 2021-08-29
fix(XMLSerializer): Preserve whitespace character references

### LINKED ISSUE #284: Incorrect serialization of XML attributes containing white space

karfau:
- Added normalization of white space literal characters as specified in https://www.w3.org/TR/xml/#AVNormalize, they now appear as space(s) in the attribute values when accessed from the DOM (and are also serialized as space)
- Added serialization of white space literal values to to the correct numerical character references as described in https://github.com/w3c/DOM-Parsing/issues/59 to fix #284 

BREAKING CHANGE: If you relied on the not spec compliant preservation of `\t`, `\n` or `\r` in **attribute values**. To preserve those you will have to create XML that instead contains the correct numerical (or hexadecimal) equivalent (`&#x9;`, `&#xA;`, `&#xD;`).

### DIFF
--- a/lib/dom-parser.js
+++ b/lib/dom-parser.js
@@ -13,15 +13,15 @@ var XMLReader = sax.XMLReader;
 /**
  * Normalizes line ending according to https://www.w3.org/TR/xml11/#sec-line-ends:
  *
- * XML parsed entities are often stored in computer files which, 
- * for editing convenience, are organized into lines. 
- * These lines are typically separated by some combination 
+ * XML parsed entities are often stored in computer files which,
+ * for editing convenience, are organized into lines.
+ * These lines are typically separated by some combination
  * of the characters CARRIAGE RETURN (#xD) and LINE FEED (#xA).
- * 
- * To simplify the tasks of applications, the XML processor must behave 
- * as if it normalized all line breaks in external parsed entities (including the document entity) 
+ *
+ * To simplify the tasks of applications, the XML processor must behave
+ * as if it normalized all line breaks in external parsed entities (including the document entity)
  * on input, before parsing, by translating all of the following to a single #xA character:
- * 
+ *
  * 1. the two-character sequence #xD #xA
  * 2. the two-character sequence #xD #x85
  * 3. the single character #x85
--- a/lib/dom-parser.js
+++ b/lib/dom-parser.js
@@ -13,20 +13,20 @@ var XMLReader = sax.XMLReader;
 /**
  * Normalizes line ending according to https://www.w3.org/TR/xml11/#sec-line-ends:
  *
- * XML parsed entities are often stored in computer files which,
- * for editing convenience, are organized into lines.
- * These lines are typically separated by some combination
- * of the characters CARRIAGE RETURN (#xD) and LINE FEED (#xA).
- *
- * To simplify the tasks of applications, the XML processor must behave
- * as if it normalized all line breaks in external parsed entities (including the document entity)
- * on input, before parsing, by translating all of the following to a single #xA character:
- *
- * 1. the two-character sequence #xD #xA
- * 2. the two-character sequence #xD #x85
- * 3. the single character #x85
- * 4. the single character #x2028
- * 5. any #xD character that is not immediately followed by #xA or #x85.
+ * > XML parsed entities are often stored in computer files which,
+ * > for editing convenience, are organized into lines.
+ * > These lines are typically separated by some combination
+ * > of the characters CARRIAGE RETURN (#xD) and LINE FEED (#xA).
+ * >
+ * > To simplify the tasks of applications, the XML processor must behave
+ * > as if it normalized all line breaks in external parsed entities (including the document entity)
+ * > on input, before parsing, by translating all of the following to a single #xA character:
+ * >
+ * > 1. the two-character sequence #xD #xA
+ * > 2. the two-character sequence #xD #x85
+ * > 3. the single character #x85
+ * > 4. the single character #x2028
+ * > 5. any #xD character that is not immediately followed by #xA or #x85.
  *
  * @param {string} input
  * @returns {string}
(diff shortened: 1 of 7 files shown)


brodybits:
I have a couple of questions:

- I noticed a reference to https://www.w3.org/TR/xml/#AVNormalize in the changes to `lib/sax.js` but not in the changes to `lib/dom.js`, wondering if it should be in both.
- I wonder if reference to https://www.w3.org/TR/xml/#AVNormalize should be in `test/parse/parse-element.test.js` as well?

I am also wondering if xmltest can eventually be considered a total overlap with the tests we have added and fixed, or if we will just keep it as an extra test in the long term.

brodybits (approved):
LGTM, thanks.

It would be nice to get my questions answered but I would consider them to be non-blocking.

karfau:
> I am also wondering if xmltest can eventually be considered a total overlap with the tests we have added and fixed, or if we will just keep it as an extra test in the long term.

My perspective on xmltest is that it provides an orientation regarding going into the "right"/expected direction (actual getting closer to or same as expected) and I would prefer to keep this, even if it's treated as a separate test run (e.g. only after the other tests passed).
The drawback is that they are not very good at communicating why something is expected.

The more I work with all the tests, the clearer the redundancies get, despite all the chaos. I think at some point not to far in the future we will be able to unclutter this without loosing any coverage/mutation score.

Regarding comments in lib, I usually put them to make things more obvious.
The git history always provides more information on details. But I'll check.

lib/dom.js
```diff
@@ -1142,7 +1142,7 @@ function needNamespaceDefine(node, isHTML, visibleNamespaces) {
  * @see https://www.w3.org/TR/xml/#NT-AttValue
```
karfau, line 1142:
```suggestion
  * @see https://www.w3.org/TR/xml/#NT-AttValue
  * @see https://www.w3.org/TR/xml/#AVNormalize
```

test/parse/parse-element.test.js
```diff
@@ -72,6 +73,46 @@ describe('XML Node Parse', () => {
 			})
 		})
 
+		describe('containing whitespace', () => {
```
karfau, line 76:
```suggestion
  // https://www.w3.org/TR/xml/#AVNormalize
		describe('containing whitespace', () => {
```

karfau (commented):
Your suggestions for additional comments make total sense

lib/dom.js
```diff
@@ -1139,10 +1139,11 @@ function needNamespaceDefine(node, isHTML, visibleNamespaces) {
  * Well-formed constraint: No < in Attribute Values
  * The replacement text of any entity referred to directly or indirectly in an attribute value must not contain a <.
  * @see https://www.w3.org/TR/xml/#CleanAttrVals
- * @see https://www.w3.org/TR/xml/#NT-AttValue
+  * @see https://www.w3.org/TR/xml/#NT-AttValue
+  * @see https://www.w3.org/TR/xml/#AVNormalize
```
brodybits, line 1143: nit: consistent indenting

```suggestion
 * @see https://www.w3.org/TR/xml/#NT-AttValue
 * @see https://www.w3.org/TR/xml/#AVNormalize
```

lib/dom.js
```diff
@@ -1139,10 +1139,11 @@ function needNamespaceDefine(node, isHTML, visibleNamespaces) {
...
  */
 function addSerializedAttribute(buf, qualifiedName, value) {
-	buf.push(' ', qualifiedName, '="', value.replace(/[<&"]/g,_xmlEncoder), '"')
+	buf.push(' ', qualifiedName, '="', value.replace(/[<&"\t\n\r]/g,_xmlEncoder), '"')
```
brodybits, line 1146: nit: add a missing space here

```suggestion
	buf.push(' ', qualifiedName, '="', value.replace(/[<&"\t\n\r]/g, _xmlEncoder), '"')
```

brodybits (approved):
LGTM with a couple more nits. Yeah your comments make total sense.

lib/dom-parser.js
```diff
@@ -13,15 +13,15 @@ var XMLReader = sax.XMLReader;
 /**
  * Normalizes line ending according to https://www.w3.org/TR/xml11/#sec-line-ends:
  *
- * XML parsed entities are often stored in computer files which, 
- * for editing convenience, are organized into lines. 
- * These lines are typically separated by some combination 
+ * XML parsed entities are often stored in computer files which,
+ * for editing convenience, are organized into lines.
+ * These lines are typically separated by some combination
  * of the characters CARRIAGE RETURN (#xD) and LINE FEED (#xA).
```
brodybits, line 19: Shouldn't these lines which are quoted start with `>`, just like we did for `Well-formed constraint: No < in Attribute Values` in `lib/dom.js`?

lib/dom.js
```diff
@@ -1137,12 +1137,18 @@ function needNamespaceDefine(node, isHTML, visibleNamespaces) {
 }
 /**
  * Well-formed constraint: No < in Attribute Values
- * The replacement text of any entity referred to directly or indirectly in an attribute value must not contain a <.
- * @see https://www.w3.org/TR/xml/#CleanAttrVals
- * @see https://www.w3.org/TR/xml/#NT-AttValue
+ * > The replacement text of any entity referred to directly or indirectly
+ * > in an attribute value must not contain a <.
```
brodybits, line 1141: +1 for starting quoted lines with `>` here 👍

brodybits (commented):
Still LGTM with a minor comment / question

Issue #271, opened 2021-07-28
[Security] now published as @xmldom/xmldom


karfau:
# &TLDR;

We published 0.7.0 to npm as [`@xmldom/xmldom`](https://www.npmjs.com/package/@xmldom/xmldom) and will continue to publish updates there.
To get the security update, you have to switch the package you use in your `package.json`.

["quick post-mortem" by @brodybits](https://github.com/xmldom/xmldom/issues/271#issuecomment-902249697)

# Original Summary

To update the library to the newest version including the latest security fix you will have to [install to from the github repo or download the artifact from the github release and install it locally](https://github.com/xmldom/xmldom/discussions/270#discussioncomment-1080132).

For details of how to do that and asking questions please use [the related discussion](https://github.com/xmldom/xmldom/discussions/270)

We have filed a ticket at the npm support team addressing the issue.

We will Post updates about the current status here but lock this issue to only allow additions by maintainers, to allow people to subscribe and get informed when something changes.

karfau:
> GitHub (GitHub Support)
> Aug 5, 2021, 1:39 PM UTC
> 
> Christian,
> 
> We are investigating your issue.
> 
> Thank you for your patience,
> npm engineering

---

>  GitHub (GitHub Support)
> Aug 5, 2021, 6:25 PM UTC
> Hi Christian,
> 
> Thanks for reaching out.
> 
> We are currently unable to find any documentation or transactions, to confirm the unscoped xmldom package was previously owned by `@xmldom` org.
> 
> Please forward any supporting document or communication for the package ownership. Once received, we'll review your request to transfer the package to xmldom org.
> 
> Best.
> 
> Judith
> GitHub Support

---

I replied today (6th of August)

Hello Judith,

I just checked `npm info` for the different xmldom versions:
The members of the npm xmldom org released all versions since 0.1.29.

The last release that was published by the maintainers listed on the npm website (jinjinyun and jindw) was 0.1.27 in November 2016

We documented the transition in our [changelog](https://github.com/xmldom/xmldom/blob/master/CHANGELOG.md#maintainer-changes), which is shipped with the packages since 0.3.0 (the linked section since 0.4.0).

Just in case this isn't enough, I Cced Chris Brody so he can reply with some reference of the mail conversation that lead to npm providing access to the package back then, in case it's still available.


Best,
Christian

karfau:
Just to clarify that there was a misunderstanding from my side, I'm adding here the reply that @brodybits  added a day after my response:

[...]

PR #311, opened 2021-08-29
docs(readme): Add downloads per week badges

karfau:
for both `xmldom` and `@xmldom/xmldom` since I keep opening both pages to check the progress of the transition:
- [`@xmldom/xmldom` npm downloads per week](https://www.npmjs.com/package/@xmldom/xmldom) 
- [`xmldom` npm downloads per week](https://www.npmjs.com/package/xmldom) 

I think having the badges at this spot will indicate when we will be able to remove this line from the readme.
It's worth mentioning that more than half of the downloads originates from deprecated versions of `xmldom` (this information is available under the versions tab):
image

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,6 +1,6 @@
 # @xmldom/xmldom
 
-***Since version 0.7.0 this package is published to npm as [`@xmldom/xmldom`](https://www.npmjs.com/package/@xmldom/xmldom) and no longer as [`xmldom`](https://www.npmjs.com/package/xmldom), because [we are no longer able to publish `xmldom`](https://github.com/xmldom/xmldom/issues/271).***  
+***Since version 0.7.0 this package is published to npm as [`@xmldom/xmldom` ![npm downloads per week](https://img.shields.io/npm/dw/@xmldom/xmldom?style=flat-square)](https://www.npmjs.com/package/@xmldom/xmldom) and no longer as [`xmldom` ![npm downloads per week](https://img.shields.io/npm/dw/xmldom?style=flat-square)](https://www.npmjs.com/package/xmldom) , because [we are no longer able to publish `xmldom`](https://github.com/xmldom/xmldom/issues/271).***  
 *For better readability in the docs we will continue to talk about this library as "xmldom".*
 
 [![license](https://img.shields.io/npm/l/@xmldom/xmldom?color=blue&style=flat-square)](LICENSE)


karfau:
I also checked if we could rather use number of dependents, but the badge available at shields.io displayed 0 for the scoped package even though npm shows 34. So there seems to be some delay in updating these numbets.

readme.md
```diff
@@ -1,6 +1,6 @@
 # @xmldom/xmldom
 
-***Since version 0.7.0 this package is published to npm as [`@xmldom/xmldom`](https://www.npmjs.com/package/@xmldom/xmldom) and no longer as [`xmldom`](https://www.npmjs.com/package/xmldom), because [we are no longer able to publish `xmldom`](https://github.com/xmldom/xmldom/issues/271).***  
+***Since version 0.7.0 this package is published to npm as [`@xmldom/xmldom` ![npm downloads per week](https://img.shields.io/npm/dw/@xmldom/xmldom?style=flat-square)](https://www.npmjs.com/package/@xmldom/xmldom) and no longer as [`xmldom` ![npm downloads per week](https://img.shields.io/npm/dw/xmldom?style=flat-square)](https://www.npmjs.com/package/xmldom) , because [we are no longer able to publish `xmldom`](https://github.com/xmldom/xmldom/issues/271).***
```
brodybits, line 3: nit: maybe better to add a space below:

```suggestion
***Since version 0.7.0 this package is published to npm as [`@xmldom/xmldom` ![npm downloads per week](https://img.shields.io/npm/dw/@xmldom/xmldom?style=flat-square)](https://www.npmjs.com/package/@xmldom/xmldom) and no longer as [`xmldom` ![npm downloads per week](https://img.shields.io/npm/dw/xmldom?style=flat-square)](https://www.npmjs.com/package/xmldom) , because [we are no longer able to publish `xmldom`](https://github.com/xmldom/xmldom/issues/271).***  

```

brodybits:
My general reaction is that this seems to add to the clutter in readme.md.

I would favor it if we could find a way to shorten this line instead and move some info to either the referenced issue #271 or into a new issue.

And I would say that *all* published versions of xmldom should have been deprecated due to the known security issues that we have fixed in `@xmldom/xmldom`.

karfau:
I agree regarding clutter, it's just an experiment. 

It's a great idea to add the badges to the summary of #271 since everything links there.

Since we lost publish rights we are also no longer able to deprecate those versions. But since they have been marked as having a security issue it's very similar.

Issue #252, opened 2021-06-18
Not all valid HTML attribute names can be parsed


karfau:
This is a continuation of https://github.com/jindw/xmldom/issues/267

Two examples that are used in vue, are attributes starting with `:` or `@`.

According to the XML 1.1 spec https://www.w3.org/TR/xml11/#sec-common-syn:
> The Namespaces in XML Recommendation [XML Names] assigns a meaning to names containing colon characters. Therefore, authors should not use the colon in XML names except for namespace purposes, but XML processors must accept the colon as a name character.

But according to the latest HTML spec, in which namespaces are not something to consider, the definition of what is allowed in attribute names is different: https://html.spec.whatwg.org/multipage/syntax.html#attributes-2

karfau:
We will need to understand how this would impact the current parsing of attributes with a namespace prefix.

We also need to decide whether there should be a different behavior for XML vs HTML. If there needs to be different behavior this is blocked by the upcoming solution for #203

marrus-sh:
So, to add additional context to this, the mentioned _Namespaces in XML Recommendation_ introduces a new concept known as “namespace-well-formed” which requires element and attribute names to match the `QName` production, and other names to match the `NCName` production, from that specification [XML Names](https://www.w3.org/TR/xml-names/). `QName` notably forbids leading colons in attribute names, so while `:foo` *is* a well‐formed XML Name, it does not produce namespace‐well‐formed documents.

Because XML parsing for the DOM is namespace‐aware, documents need to be namespace‐well‐formed in addition to just being ordinarily well‐formed. So it *should* produce an error when parsing as XML.

HTML’s rules have been made considerably more lax in order to support `data-*` attributes and HTML custom elements. Importantly, it is entirely possible to write an HTML custom element with a `:foo` attribute. So ideally you would have different behaviour here.

karfau:
@marrus-sh Thank you, that was very helpful. I updated my last comment accordingly.

Kevin44:
It also appears that there is a problem with the attribute "constructor" also.  Here is an example element with that attribute that it is failing on: 

<java:new doc:name="Get date instance" doc:id="f5784e1d-4a0d-428f-9833-9f33694bbf10" class="java.util.Date" constructor="Date()" target="dateVar"/>

marrus-sh:
[...]

PR #315, opened 2021-09-01
fix(sax): Restore ability to parse `__prototype__` attributes

dsimpsonOMF:
As noted in issue #252 some attributes conflict with prototype fields (such as an attribute `constructor`) thus we must remove the `in` operator and replace with `Object.prototype.hasOwnProperty()`.


karfau (approved):
Thank you @dsimpsonOMF

karfau:
was released as 0.7.4

PR #317, opened 2021-09-04
test: Add executable examples for node and typescript

karfau:
In the context of the threads in #214 I'm have added executable examples and test them as github actions to make sure they don't break.

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -1,0 +1,32 @@
+# This workflow will run the examples provided in the examples folder
+name: examples
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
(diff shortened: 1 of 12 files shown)


brodybits:
I will try to review later this week, apologies for the extra delays.

karfau:
@brodybits Thx, but even though I'm "happy about the current state of this PR", which is why it's no longer a draft, it is just an experiment.
I would prefer if you can look into the open topics related to the [0.8.0 milestone](https://github.com/xmldom/xmldom/milestone/10)

karfau:
@brodybits please let me know if you have time to look at this. I would love to land this before continuing to work on #338, to see the impact on potential changes to the type definitions.

If you don't have time for this, just let meknow as well, in that case I'm confident enough to land it.

PR #321, opened 2021-09-09
fix(dom): Preserve default namespace when serializing

### LINKED ISSUE #319: Regression in 0.7.0: Namespace prefixes being introduced unnecessarily
We are using xmldom to parse, tweak, and serialize some SVG schematics that will ultimately be embedded in HTML (not xhtml) pages. Through 0.6.0, everything was working fine, but starting with 0.7.0, for some but not all documents, when we SerializeToString(doc), all of the elements in the document get prefixed by the svg namespace (<xvg:g>, etc.) This produces SVG documents that are perfectly legitimate XML but won't display when embedded in HTML5 documents due to HTML rules, which don't include the general XML rules for namespaces. 

1. What causes xmldom to do this? We didn't discover an identifiable pattern. 
2. How can we ensure that we get SVG we can use?

Right now we're hacking the results as text to lop off the namespaces so our app will continue to work, but we really don't want to do that.

Update by @karfau:
The regression was introduced in #268

lupestro:
The regression was introduced as part of #268 by not considering the visible default namespace for attributes without a prefix.

Fixes #319

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -488,6 +488,20 @@ Node.prototype = {
     hasAttributes:function(){
     	return this.attributes.length>0;
     },
+	/**
+	 * Look up the prefix associated to the given namespace URI, starting from this node.
+	 * **The default namespace declarations are ignored by this method.**
+	 * See Namespace Prefix Lookup for details on the algorithm used by this method.
+	 *
+	 * _Note: The implementation seems to be incomplete when compared to the algorithm described in the specs._
+	 *
+	 * @param {string | null} namespaceURI
+	 * @returns {string | null}
+	 * @see https://www.w3.org/TR/DOM-Level-3-Core/core.html#Node3-lookupNamespacePrefix
+	 * @see https://www.w3.org/TR/DOM-Level-3-Core/namespaces-algorithms.html#lookupNamespacePrefixAlgo
+	 * @see https://dom.spec.whatwg.org/#dom-node-lookupprefix
+	 * @see https://github.com/xmldom/xmldom/issues/322
+	 */
     lookupPrefix:function(namespaceURI){
     	var el = this;
     	while(el){
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -1175,10 +1175,21 @@ function serializeToString(node,buf,isHTML,nodeFilter,visibleNamespaces){
 		var prefixedNodeName = nodeName
 		if (!isHTML && !node.prefix && node.namespaceURI) {
 			var defaultNS
+			// lookup current default ns from `xmlns` attribute
 			for (var ai = 0; ai < attrs.length; ai++) {
 				if (attrs.item(ai).name === 'xmlns') {
 					defaultNS = attrs.item(ai).value
 					break
+				}
+			}
+			if (!defaultNS) {
+				// lookup current default ns in visibleNamespaces
+				for (var nsi = visibleNamespaces.length - 1; nsi >= 0; nsi--) {
+					var namespace = visibleNamespaces[nsi]
+					if (namespace.prefix === '' && namespace.namespace === node.namespaceURI) {
+						defaultNS = namespace.namespace
+						break
+					}
 				}
 			}
 			if (defaultNS !== node.namespaceURI) {
--- a/test/dom/serializer.test.js
+++ b/test/dom/serializer.test.js
@@ -119,6 +119,33 @@ describe('XML Serializer', () => {
 			)
 		})
 	})
+	describe('is insensitive to namespace order', () => {
+		it('should produce unprefixed svg elements regardless of xmlns vs xmnls:svg order', () => {
+			const svgNamedIsFirst = [
+				'<svg version="1.1" width="153" height="144" ',
+				'  xmlns:svg="http://www.w3.org/2000/svg" ',
+				'  xmlns="http://www.w3.org/2000/svg" >',
+				'<g><circle cx="60" cy="60" r="50"/></g>',
+				'</svg>',
+			].join('\n')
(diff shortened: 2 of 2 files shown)


karfau:
Thx again for investing the time :)

We will release 0.7.5 as soon as this fix is ready. Support is very welcome, here are my thoughts:
1. I think we should also add tests for correct treatment of "overlapping namespaces" when they are overlapping on different levels:

   - default namepsace on parent & prefixed in nested => no prefixes at all
   - prefixed namespace on parent & default namespace in nested => prefixes only when default one is not present
to be sure those keep working as expected. 

I think the core question here is whether the prefix would always change to the default one if it's available, or if it should just be preserved as it was before. Again it would be a good idea to run this check in a browser and also against 0.6.0.

Here are my steps to check those tests against 0.6.0:
1. manually change the `describe` you added to be `describe.only`
2. execute the following:
```bash
git checkout 0.6.0 -- lib
npm test -- test/dom/serializer.test.js
```
(To get back to current source code version use `git checkout HEAD -- lib`. Adn don't forget to remove the `.only` from the describe.)

Assuming 0.6.0/browsers always preserve the prefix, we should change the namespace resultion accordingly.

karfau:
I pushed some minor changes to the tests and placeholders for the tests I would like to be added (either by me or by you). I will think about a good example that we can use for those cases.

I'm thinking about temporarily add checkout 0.6.0 of lib and `describe.only` to this branch, as long as we are anyways researching the previous behavior.
What do you think?

lupestro:
I ran the tests against 0.6.0 and adjusted the xml spacing so that the output matches the input, since we're looking for perfect match now rather than specific features. (Newlines between `<` and `>` and extra spaces before `>` or `/>` disappear on a round trip.) Now the tests pass on 0.6.0 but fail on latest because of the `<svg:g>` and `<svg:circle>` introductions, so it is demonstrating the issue again (but much more graphically than it was before.)

As far as adjustments to the branch, I'm not sure how you do that but it sounds good to me.

karfau:
@lupestro Thx for the nice collaboration.
I added the tests I was looking for. I'm glad it's really just the serialization that is affected and not the parsing.

I'll call it a day and the next step will be to think about the proper solution.

We do not run the tests in any browsers so far, I have thought about this, but didn't invest time into any research of possible tools yet.

karfau:
I dug deeper into what changes caused this issue and now I'm confident that it was this one:

https://github.com/xmldom/xmldom/commit/c0fc02a289ca16a4aad2477b600b71d335380ab6#diff-79db8c65cf49002eb2d5d7e81a17f3b73d94db6bb50c0c12f0c9f427aff4b11eR1076-R1092

For unprefixed attributes the default namespace is only checked on the level of attributes, but not on the level of visible namespaces.
I think it will not take me long to fix it.

karfau:
Although I found the general fix, I agree that at least some part of the code I linked and added chould actually use `lookupPrefix` instead.

I will check the definition for that method in the different specs, to see how it matches what we have in those two places.

karfau:
@lupestro The current state would be what I would land, can you test this version against your codebase (e.g. by installing from github using the commit hash `@xmldom/xmldom@github:xmldom/xmldom#9df8562`) to check that we covered all the cases (and if not, add another failing test)?
Be aware that this PR targets the default branch which already contains some [potentially breaking changes for the upcoming 0.8.0 release](https://github.com/xmldom/xmldom/pulls?q=is%3Apr+is%3Amerged+milestone%3A0.8.0+label%3A%22breaking+change%22+). If you need a version based on 0.7 I can create that and post the related commit hash here.

I will give other maintainers a change to look into this until Monday evening, after which I will release it as 0.7.5.

lupestro:
Tried it with our stuff and it's now behaving correctly. So green light from our end.

lib/dom.js
```diff
@@ -1175,12 +1189,23 @@ function serializeToString(node,buf,isHTML,nodeFilter,visibleNamespaces){
 		var prefixedNodeName = nodeName
 		if (!isHTML && !node.prefix && node.namespaceURI) {
 			var defaultNS
+			// lookup current default ns from `xmlns` attribute
```
brodybits, line 1192: minor nit: I wonder if we could somehow use Array.prototype.find() here, assuming we can start dropping some extra-old browsers?

lib/dom.js
```diff
@@ -1175,12 +1189,23 @@ function serializeToString(node,buf,isHTML,nodeFilter,visibleNamespaces){
...
 			for (var ai = 0; ai < attrs.length; ai++) {
 				if (attrs.item(ai).name === 'xmlns') {
 					defaultNS = attrs.item(ai).value
 					break
 				}
 			}
+			if (!defaultNS) {
+				// lookup current default ns in visibleNamespaces
+				for (var nsi = visibleNamespaces.length - 1; nsi >= 0; nsi--) {
```
brodybits, line 1201: (minor) nit / question: could we replace some of this code with this kind of pattern:

    Array.from(visibleNamespaces).reverse().find(...)

brodybits (approved):
LGTM overall, thanks to both @lupestro & @karfau

I left a couple minor nits which should probably be left for future consideration (maybe).

lib/dom.js
karfau: Of course it would be less code, but with additional memory cost. And since this can happen quite often, I would not consider it.

lib/dom.js
karfau: Currently we support ES5, requiring Array.find to be available would be a massive drop of supported (old) versions as far as I can judge. And I'm not aware how could measure/judge when to do that.
I think we should discuss how and when to drop support (e.g. in the related runtimes discussion). But not "now".

PR #398, opened 2022-04-06
feat: Add traversal methods to Element

KevinYeramian:
What does this PR do? 
- Add traversal methods to `Element`

I used the word `traversal` because it used in the official documentation to refers to these methods ([link](https://www.w3.org/TR/ElementTraversal/#attribute-lastElementChild))

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -636,6 +636,25 @@ function _onUpdateChild (doc, el, newChild) {
  * @private
  */
 function _removeChild (parentNode, child) {
+	if (child.nodeType === ELEMENT_NODE) {
+		parentNode.childElementCount -= 1;
+
+		if (child.previousElementSibling) {
+			child.previousElementSibling.nextElementSibling = child.nextElementSibling;
+		} else {
+			parentNode.firstElementChild = child.nextElementSibling;
+		}
+
+		if (child.nextElementSibling) {
+			child.nextElementSibling.previousElementSibling = child.previousElementSibling;
+		} else {
+			parentNode.lastElementChild = child.previousElementSibling;
+		}
+
+		child.previousElementSibling = null;
+		child.nextElementSibling = null;
+	}
+
 	var previous = child.previousSibling;
 	var next = child.nextSibling;
 	if (previous) {
@@ -721,6 +740,19 @@ function _appendSingleChild (parentNode, newChild) {
 	} else {
 		parentNode.firstChild = newChild;
 	}
+
+	if (newChild.nodeType === ELEMENT_NODE) {
+		newChild.previousElementSibling = parentNode.lastElementChild
+		if (newChild.previousElementSibling) {
+			parentNode.lastElementChild.nextElementSibling = newChild
+		} else {
+			parentNode.firstElementChild = newChild;
+		}
+
+		parentNode.lastElementChild = newChild
+		parentNode.childElementCount += 1
+	}
+
 	parentNode.lastChild = newChild;
 	_onUpdateChild(parentNode.ownerDocument, parentNode, newChild);
 	return newChild;
@@ -831,7 +863,6 @@ Document.prototype = {
 		node.nodeName = tagName;
 		node.tagName = tagName;
 		node.localName = tagName;
-		node.childNodes = new NodeList();
 		var attrs	= node.attributes = new NamedNodeMap();
 		attrs._ownerElement = node;
 		return node;
@@ -887,7 +918,6 @@ Document.prototype = {
 		var node = new Element();
 		var pl = qualifiedName.split(':');
 		var attrs	= node.attributes = new NamedNodeMap();
-		node.childNodes = new NodeList();
 		node.ownerDocument = this;
 		node.nodeName = qualifiedName;
 		node.tagName = qualifiedName;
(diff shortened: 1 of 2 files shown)


test/dom/element.test.js
```diff
@@ -208,6 +208,87 @@ describe('Document', () => {
 		expect(doc.childNodes.toString()).toBe(`<A/><B/><C/>`)
 	})
 
+	it('ElementTraversal', () => {
```
karfau, line 211: I didn't dig into every line to understand it, but I would prefer to have multiple `it`s that describe "independent" behaviour/scenarios and if some things only make sense to be tested together, of course leave them together. But my guess is that your comments might indicate some "sections"?

I don't have an issue with nested `describe` blocks, so maybe it's like 
```suggestion
	describe('element traversal', () => {
```
and it contains the different scenarios as `it`s?

lib/dom.js
```diff
@@ -831,7 +863,6 @@ Document.prototype = {
 		node.nodeName = tagName;
 		node.tagName = tagName;
 		node.localName = tagName;
-		node.childNodes = new NodeList();
```
karfau, line 834: That makes a lot of sense. But is there a specific reason to move it as part of this PR?
If not, would you mind to file a separate PR for just this refactoring ...

lib/dom.js
```diff
@@ -831,7 +863,6 @@ Document.prototype = {
...
 		var attrs	= node.attributes = new NamedNodeMap();
 		attrs._ownerElement = node;
```
karfau, line 867: ... including moving this to the Element constructor?
(The same is true for `createElementNS`)

karfau (commented):
The method [`_insertBefore`](https://github.com/xmldom/xmldom/pull/398/files#diff-79db8c65cf49002eb2d5d7e81a17f3b73d94db6bb50c0c12f0c9f427aff4b11eR679) is calling the remove method but has extra code for joining inserting a child into the NodeList, can you please check if something also needs to be added there?

karfau:
We also need some documentation:
- listing them in the readme as part of the section of the spec that first introduced them.
- jsdoc comments on the properties with minimal description taken from the spec, pointing to any differences from the spec.
And ideally I would prefer to mark those properties as `@readonly` in a "tsdoc" comment way (you should be able to find other examples in the code base), to get a bit closer to having getters instead of normal properties.

karfau:
Thank you a lot for the contribution so far.
I provided some feedback that needs to be addressed, but it seems to be a much smaller effort than I originally anticipated.

KevinYeramian:
Review fixes committed!
- I am not sure what to check for your [comment](https://github.com/xmldom/xmldom/pull/398#pullrequestreview-934124149)
- I added the "tsdoc" but it current format seems little bit “heavy" in the code.

readme.md
```diff
@@ -190,6 +190,10 @@ import { DOMParser } from '@xmldom/xmldom'
 		
   readonly attribute:
   - `tagName`
+  - `firstElementChild`
+  - `lastElementChild`
+  - `nextElementSibling`
+  - `previousElementSibling`
```
karfau, line 196: What about `childElementCount`?

test/dom/element.test.js
```diff
@@ -208,6 +208,79 @@ describe('Document', () => {
...
+
+		it('append one element and remove', () => {
+			const doc = new DOMParser().parseFromString(`<A></A>`)
+			const aElement = doc.firstChild
+			const bElement = doc.createElement('B')
+			aElement.appendChild(bElement)
+
+			expect(aElement.childElementCount).toBe(1)
+			expect(aElement.firstElementChild === bElement).toBe(true)
+			expect(aElement.lastElementChild === bElement).toBe(true)
```
karfau, line 231: Minor: This could be written as
```suggestion
			expect(aElement.firstElementChild).toBe(bElement)
			expect(aElement.lastElementChild).toBe(aElement.firstElementChild)
```
What do you think about this approach?

karfau:
> Review fixes committed!

Nice, loving it so far

> * Not sure what to check for your [comment](https://github.com/xmldom/xmldom/pull/398#pullrequestreview-934124149)

Yeah, that comment is a bit hard to grasp, sorry.
I think we are missing updating the new props as part of the `Node.insertBefore` API, which is also used as part of other methods:
https://github.com/brodybits/xmldom/blob/c234c4d08b482841886b703547dd62697e6d863d/lib/dom.js#L448-L462

> * I added the "tsdoc" but it current format seems little bit “heavy" in the code.

Looks great for me.

karfau:
> The method [`_insertBefore`](https://github.com/xmldom/xmldom/pull/398/files#diff-79db8c65cf49002eb2d5d7e81a17f3b73d94db6bb50c0c12f0c9f427aff4b11eR679) is calling the remove method but has extra code for joining inserting a child into the NodeList, can you please check if something also needs to be added there?

@KevinYeramian This comment points to my main concern for this PR so far. Do you still have capacity to continue working on it? I can help torelsove the conflict if that's an issue.

karfau:
I resolved the commit, which leads to failing tests (due to the breaking changes in version 9.
I will see if I find time to get back to it, but it just didn't feel right to close the PR.

Issue #44, opened 2020-04-07
Adding elements then serializing results in non-formatted text


Benjin:
Parsing this XML:
```
<Parent>
    <FirstChild>
        <Grandchild />
    </FirstChild>
</Parent>
```

then appending a new child element to `Parent`, then serializing back to text results in this:

```
<Parent>
    <FirstChild>
        <Grandchild />
    </FirstChild>
<SecondChild><Grandchild /><SecondChild></Parent>
```

instead of the expected:

```
<Parent>
    <FirstChild>
        <Grandchild />
    </FirstChild>
    <SecondChild>
        <Grandchild />
    <SecondChild>
</Parent>
```

Is this expected?  Is there an option I can set somewhere to ensure serialized XML is human-readable/"pretty"?

karfau:
Thank you for filing this and sorry that we have not been able to reply earlier, it must have slipped through somehow.

There is currently no option for pretty printing as part of the serialization, but the existing formatting of the input is preserved by also serializing the text nodes between the different elements. You could of course append the expected "text nodes" between the newly added elements to do something like pretty printing.

I'm not sure if it would be outside of the scope of this library to implement something like that, we should check how browser are treating this case.

At least we should come to a conclusion and properly document this before we reach 1.0

Benjin:
My vote is for a pretty-print option!  Formatted XML is far easier to PR, so it's a common-enough case that saving me from having to find another library to handle that part of the serialization would be nice.

karfau:
If you have to PR things then it's about stuff that you check into version control, so you could use one of the many tools available to prettify source code.
There are many different solutions possible, just to name a few:
- applying an XSLT transformation
- implementing a visitor that aligns text nodes in the tree so that nested elements are indented

They all boil down to inserting text nodes containing the required white space before or during serializing.

There are quite some things to consider like "what white space to use to indent?", "how to treat existing white space that doesn't match the space that should be used to use to indent?", ...

I'm not aware of any of [the specs we try to align with](https://github.com/xmldom/xmldom#specs) that has any related API/feature build in. So I do not think the current maintainers will invest time into this before 1.0

Feel free to post solutions/workarounds/proposals in this issue.
If you want to even provide a PR (including tests!) that offer formatting as an option, be aware that we might still reject it if we think we will not be able to maintain it.

eamanola:
@karfau I can make a PR for this. do you have test documents?

eamanola:
[...]

PR #404, opened 2022-04-27
#44 add pretty option to serializer

eamanola:
something like this


eamanola:
after playing around a bit, git the same white space issue. its too big a puzzle to generalize.

closing PR

PR #416, opened 2022-06-23
fix: Complete multi-character sanitization

ACN-kck:
Go to know by CodeQL in another project, that there is a missing multi-character sanitization.

https://github.com/whelk-io/maven-settings-xml-action/pull/144/files#diff-3d2b59189eeedc2d428ddd632e97658fe310f587f7cb63b01f9b98ffc11c0197R909

image

### DIFF
--- a/lib/sax.js
+++ b/lib/sax.js
@@ -120,12 +120,11 @@ function parse(source,defaultNSMapCopy,entityMap,domBuilder,errorHandler){
 				var tagName = source.substring(tagStart + 2, end).replace(/[ \t\n\r]+$/g, '');
 				var config = parseStack.pop();
 				if(end<0){
-
-	        		tagName = source.substring(tagStart+2).replace(/[\s<].*/,'');
+	        		tagName = source.substring(tagStart+2).replace(/[\s<].*/g,'');
 	        		errorHandler.error("end tag name: "+tagName+' is not complete:'+config.tagName);
 	        		end = tagStart+1+tagName.length;
 	        	}else if(tagName.match(/\s</)){
-	        		tagName = tagName.replace(/[\s<].*/,'');
+	        		tagName = tagName.replace(/[\s<].*/g,'');
 	        		errorHandler.error("end tag name: "+tagName+' maybe not complete');
 	        		end = tagStart+1+tagName.length;
 				}
--- a/lib/sax.js
+++ b/lib/sax.js
@@ -124,9 +124,9 @@ function parse(source,defaultNSMapCopy,entityMap,domBuilder,errorHandler){
 	        		errorHandler.error("end tag name: "+tagName+' is not complete:'+config.tagName);
 	        		end = tagStart+1+tagName.length;
 	        	}else if(tagName.match(/\s</)){
+	        		end = tagStart + 2 + tagName.length;
 	        		tagName = tagName.replace(/[\s<].*/g,'');
 	        		errorHandler.error("end tag name: "+tagName+' maybe not complete');
-	        		end = tagStart+1+tagName.length;
 				}
 				var localNSMap = config.localNSMap;
 				var endMatch = config.tagName == tagName;
--- a/test/parse/__snapshots__/simple.test.js.snap
+++ b/test/parse/__snapshots__/simple.test.js.snap
@@ -29,7 +29,7 @@ Object {
 
 exports[`parse unclosed inner 1`] = `
 Object {
-  "actual": "<r><Page><Label/></Page>  <Page/></r>",
+  "actual": "<r><Page><Label/></Page></r>",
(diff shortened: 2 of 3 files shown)


karfau:
Thank you for the report and for the PR.
I was able to verify the bug:
when parsing the following XML:
`<xml><child></child <injected></xml>`
xmldom produces the dom equivalent to
`<xml><child /><injected /></xml>`
xmldom also reports at least one error in this situation, so not trusting any xml that reports errors (by adding a custom error handler, see https://github.com/xmldom/xmldom#api-reference ) is the best workaround currently.

Sadly the PR doesn't fix that (potential security) issue, which is not documented publicly.

ACN-kck:
> Thank you for the report and for the PR. I was able to verify the bug: when parsing the following XML: `<xml><child></child <injected></xml>` xmldom produces the dom equivalent to `<xml><child /><injected /></xml>` xmldom also reports at least one error in this situation `[xmldom error] end tag name: child maybe not complete`, so not trusting any xml that reports errors (by adding a custom error handler, see https://github.com/xmldom/xmldom#api-reference ) is the best workaround currently.
> 
> Sadly the PR doesn't fix that (potential security) issue, which is not documented publicly.

I reviewed my solution again which indeed does not fix the issue. I checked again how this could be resolved by creating one additional test based on your message.

With your given example:
`<xml><child></child <injected></xml>` should be parsed to
`<xml><child/></xml>`, correct?

lib/sax.js
```diff
@@ -120,12 +120,11 @@ function parse(source,defaultNSMapCopy,entityMap,domBuilder,errorHandler){
...
 				var config = parseStack.pop();
 				if(end<0){
-
-	        		tagName = source.substring(tagStart+2).replace(/[\s<].*/,'');
+	        		tagName = source.substring(tagStart+2).replace(/[\s<].*/g,'');
 	        		errorHandler.error("end tag name: "+tagName+' is not complete:'+config.tagName);
 	        		end = tagStart+1+tagName.length;
 	        	}else if(tagName.match(/\s</)){
-	        		tagName = tagName.replace(/[\s<].*/,'');
+	        		tagName = tagName.replace(/[\s<].*/g,'');
```
ACN-kck, line 127: By calculating the length first, the injection could be prevented.
Snapshot test 'unclosed inner' in simple.test.js will start failing, but while looking into it I'm struggling to understand why the test should work without errors. My expectation would be that it fails with an error.

```suggestion
	        		end = tagStart + 2 + tagName.length;
	        		tagName = tagName.replace(/[\s<].*/g,'');
```

ACN-kck: And we need to shift the index +1 to catch the closing '>'

karfau:
Yes, I agree that this support for automatically closing unclosed tags is the culprit.
So the fix will be a breaking change in the semver sense.

ACN-kck:
> Yes, I agree that this support for automatically closing unclosed tags is the culprit. So the fix will be a breaking change in the semver sense.

I'm not having the full view on the lib in regards to this breaking change.
Is there something I should have a closed look?

karfau:
Sorry for not replying for such a long time.
I will get back to looking at this PR starting from November.

mgerlach:
Would be good to have this released soon for https://github.com/whelk-io/maven-settings-xml-action/pull/144

karfau:
@mgerlach @ACN-kck what is the reason for not using tge most recent version 0.8.2 in the linked PR?

If the effort is not to high it would be awesome to have that CodeQL check as part of the xmldom pipeline.
Are you able to help me setting this up?
This could be done in a separate PR, in which case it would contribute to hacktoberfest, in case that is relevant for anybody.
And I'm going to look into the option of including it into the 0.9.0 release (I recently published a pre-release 0.9.0-beta.1 in preparation of that).

karfau:
I have tried to understand the current state of the PR, but I'm not able to relate the code suggestion in the comment to the code that was already commited.

I'm also not able to push updates to your branch, I guess you didn't check the box "allow edits by maintainers"?

So it looks like I have to replicate you changes in a local branch to better understand them.

mgerlach:
Hi @karfau I only jumped onto this because GitHub Actions is complaining about https://github.com/whelk-io/maven-settings-xml-action still using Node.js 12. https://github.com/whelk-io/maven-settings-xml-action/pull/144 would fix that (among other things). I don't know why the PR has xmldom-0.7.0 rather than 0.8.3... @ACN-kck can help @karfau please?

karfau:
@ACN-kck @mgerlach I fixed this in a more robust way in #445, which will be part of the 0.9.0 release.
I was also able to configure CodeQL checks to the repo as part of #444 

Thank you for your contributions.

PR #445, opened 2022-10-18
fix: Stop adding tags after incomplete closing tag

karfau:
Previously it was tested as a feature, that tags after incomplete closing tags would be part of the dom. The code responsible for this triggered a CodeQL error suspecting that it could lead to injection of a `script` tag.

BREAKING CHANGE: It no longer reports an error when parsing HTML containing incomplete closing tags, to align the behavior with the one in the browser.

BREAKING CHANGE: If your code relied on not well formed XML to be parsed and include subsequent tags, this will no longer work.

https://github.com/xmldom/xmldom/pull/416


ACN-kck:
Great stuff!

PR #482, opened 2023-03-19
fix: several issues in NamedNodeMap and Element

### LINKED ISSUE #46: getAttribute() doesn't return null for not existing attributes
`getAttribute()` and related methods don't return null for not existing attributes.
Methods below should be fixed to return null.

* NamedNodeMap.prototype.getNamedItem()
* Element.prototype.getAttribute()
* Element.prototype.getAttributeNode()
* Element.prototype.getAttributeNS()

karfau:
- proper order in finding attributes (only affects duplicate attributes, which can only be created manually and anyways produce invalid XML)
- properly remove item from previous last index (was not iterated over but the reference was kept)
- set `ownerElement` to `null` even when `ownerElement.ownerDocument` is not set (which is an edge case which doesn't happen when parsing XML)
- throw `DOMException` when there is no matching attribute
- add test coverage
- add doc comments regarding differences to the specification that still exist:
  - no named property indices support
- add validation of qualified names in `Element` API which can throw `DOMException` as declared in the specs
- always initialize required properties with `null` in `Attr` constructor

BREAKING CHANGES: Mainly not throwing when removing an attribute that doesn't exist and other behavior that was different from the specification.

closes #46
see #477


karfau:
According to the spec the `removeAttribute*` methods on `Element` should not throw (the same way) when an attribute is not found, but I think they would with the current implementation.
Since those methods are also not covered with tests, I will add tests for those before merging.

PR #496, opened 2023-06-11
ci: Stop running mutation tests using stryker

karfau:
since they were running for 5+ hours and because of that only on master.

Since we have quite a good code coverage now (>80% total), I would argue it's not worth all the compute invested on each master commit.

I will instead work on reporting coverage in PRs and add a badge for the master branch


karfau:
@brodybits since you introduced this, I will keep this open for some days to give you a chance to veto the change with some arguments (or a PR that speeds up the run substantially.

Issue #497, opened 2023-06-13
Trailing whitespace in processing instructions not preserved


shunkica:
**Describe the bug**
Trailing whitespace in PI is not preserved.
One trailing whitespace is preserved if the PI does not have data.

**To Reproduce**
https://stackblitz.com/edit/js-xmldom-template-mcmkty?file=index.js

**Expected behavior**
The PI without data should have no trailing whitespace.
The trailing whitespace in PI with data should be preserved.
From [DOM Level 1 spec](https://www.w3.org/TR/REC-DOM-Level-1/level-one-core.html#ID-1004215813):

> Attributes
>     target
>         The target of this processing instruction. XML defines this as being the first token following the markup that begins the processing instruction.
>     data
>         The content of this processing instruction. **This is from the first non white space character after the target to the character immediately preceding the ?>.** 

**Runtime & Version:**
xmldom version: 0.8.8

karfau:
Thank you for reporting the issue.
I think the culprit is in the regular expression in https://github.com/xmldom/xmldom/blob/master/lib/sax.js#L611, which seems to drop whitespace before the closing symbols.

Here is the same section in the DOM Level 2 spec: https://www.w3.org/TR/DOM-Level-2-Core/#core-ID-1004215813

But I still have to find the related part in the XML spec.

karfau:
If I have to tackle it myself it will happen after I released 0.9.0, so there will be a patch for 0.8.x but not for 0.7.x

But if somebody else takes care of it before 0.9.0 is released 0.7.x will also be patched.

PR #498, opened 2023-06-15
fix!: preserve DOCTYPE internal subset 

### LINKED ISSUE #117: To what degree should xmldom support DTD processing support for Entity/ EntityReference?

shunkica:
Add support for parsing the internal subset of a DOCTYPE and saving it as a string in `DocumentType.internalSubset`.

BREAKING CHANGE: Many documents that were previously accepted by xmldom, esecially non well-formed ones are no longer accepted. Some issues that were formerly reported as `error`s are now a `fatalError`.

fixes #117, #497 

Spec: [XML DOM L2 Core](https://www.w3.org/TR/DOM-Level-2-Core/#core-ID-412266927)


shunkica:
> Test Suites: 2 failed, 32 passed, 34 total
> Tests:       205 failed, 2 skipped, 727 passed, 934 total
> Snapshots:   205 failed, 261 passed, 466 total

shunkica:
> Test Suites: 2 failed, 32 passed, 34 total
> Tests:       201 failed, 2 skipped, 731 passed, 934 total
> Snapshots:   201 failed, 265 passed, 466 total

I am not comfortable with modifying the split function to extract the internalSubset, so I added a new function to extract the entire DOCTYPE element without the use of regex, and then catch everything inside between the first `[` and the last `]`

karfau:
Thx, I will have a closer look over the weekend.

shunkica:
Diving deep into validation of the entire doctype declaration against XML spec using this char-by-char state machine method seemed worthwhile initially, however, it became clear that my approach deviated from the established spirit and code style of this library.

Also, I have also determined that this `Array.indexOf(x) !== -1` way of checking for matches (which I always assumed would be more efficient than regex) is only faster than regex if there are a small number of characters in the array.

I will stop further changes, and appreciate any feedback to guide my next steps.

lib/sax.js
```diff
@@ -581,11 +593,27 @@ function parseDCC(source, start, domBuilder, errorHandler) {
...
-				var lastMatch = matchs[len - 1];
-				domBuilder.startDTD(name, pubid, sysid);
+
+				// Unquote pubid and sysid if necessary
+				if (pubid && /^(['"]).*\1$/.test(pubid)) {
+					pubid = pubid.substring(1, pubid.length - 1);
+				}
+				if (sysid && /^(['"]).*\1$/.test(sysid)) {
+					sysid = sysid.substring(1, sysid.length - 1);
+				}
```
karfau, line 603: This change seems to break quite some code regarding mixed quotes inside those ids. So I would recommend to not make that change.
I must admit that this is not documented well, so I will push a change to your PR that fixes the documentation part.

See https://github.com/xmldom/xmldom/blob/c4cb17d48feb2fbf1d4ac5ff96c48aa1e8117612/CHANGELOG.md#050

lib/sax.js
```diff
@@ -581,11 +593,27 @@ function parseDCC(source, start, domBuilder, errorHandler) {
...
+
+				// Extract internal subset
+				var doctype = extractDoctype(source, start);
```
karfau, line 606: When calling this code the context is already restricted to  we are inside a doctype, so it can not be a comment or anything else (see condition in line 583)

So to just get the internalSubset I think way less work is required as part of extractDocType. I would have expected it to be between the first `[` after the ids and the last `]`?

THe general approach I prefer is to first get something "minimal" working (including all required tests) and iterate over it in the next steps/PRs.
Especially "optimizations" should not make the code less readable/maintainable.

lib/sax.js
shunkica: > This change seems to break quite some code regarding mixed quotes inside those ids. So I would recommend to not make that change. I must admit that this is not documented well, so I will push a change to your PR that fixes the documentation part.
> 
> See https://github.com/xmldom/xmldom/blob/c4cb17d48feb2fbf1d4ac5ff96c48aa1e8117612/CHANGELOG.md#050

I understand the decision made at that point, even though strictly speaking the quotes are not a part of the indentifier.

lib/sax.js
shunkica: I understand. At the time I wrote that my end goal was replacing the entire part where the doctype is handled. 

I didn't just extract everything from the first `[` to the last `]` because at that point it is not known at which index the doctype declaration ends (or it it?)

I have been experimenting with regex to extract all the contents of DOCTYPE (to replace the regex in `split`) and came up with [this](https://regex101.com/r/fqt0AM/2):

`<!DOCTYPE\s+(\w+)(\s+(SYSTEM)\s+(['"])(\w+)\4|\s+(PUBLIC)\s+(['"])(\w+)\7\s+(['"])(\w+)\9)?\s*(\[(\s*(%\w+;|<([^>]*)>)\s*)*])?\s*>`

then we could get
name: group[1]
publicId: group[8]
systemId: group[10] || group[5]
internalSubset: group[11]

Obviously occurrences of `\w` would have to be replaced with namecharstart+namechar*

Do you think this is worth pursuing further?

lib/sax.js
karfau: If you want to follow that path (since you reverted my change to restore the quotes), you have to take care of fixing the tests that preserve the mixed quote cases in a different way.
Otherwise I can not merge this PR.
And I would prefer to not "fix everything in one PR", but instead have short lived PR with a well defined scope.

lib/sax.js
karfau: Yes, I understood that. And if you plan to also implement more things then just preserving the internalSubset, like parsing entities in this PR, using an extra method like this might make sense.
But if this PR just tries to preserve the `internlSubset` (which it already does), we should find the simplest possible implementation and elaborate on it in the next step/PR.

karfau: > I have been experimenting with regex to extract all the contents of DOCTYPE (to replace the regex in `split`)

I did not yet look at that in every detail yet, but I will, just needs more time.

test/xmltest/__snapshots__/valid.test.js.snap
```diff
@@ -534,32 +698,28 @@ Array [
...
+<doc>&amp;e;</doc>",
   "errors": Array [
-    Array [
-      "error",
-      "element parse error: Error: Hierarchy request error: Only one element can be added and only after doctype",
-      Object {
-        "columnNumber": 2,
-        "lineNumber": 6,
-      },
-    ],
```
karfau, line 557: The current implementation obviously also fixes several other issues: in this case it looks like a totally valid document was previously not parsable but now it is.
Great stuff.

lib/sax.js
shunkica: If you decide the regex approach is worth a shot I can dedicate more time to making it spec compliant, then the extractDoctype method will not be needed.
I was planning to use this PR as a stepping stone for a entity parser.

test/xmltest/__snapshots__/valid.test.js.snap
```diff
@@ -1161,16 +1525,13 @@ Object {
...
+<doc>&amp;e1;</doc>",
   "errors": Array [
-    Array [
-      "error",
-      "entity not found:&e2;",
-      Object {
-        "columnNumber": 6,
-        "lineNumber": 6,
-      },
-    ],
```
karfau, line 1173: Another one: it looks like the sax parser was previously parsing the `&e2;` from the second entity, now it's no longer complaining about it.
Great.

lib/sax.js
karfau: But if you change the regex the tests will tell us whether anything breaks, so give it a try, we can always revert!

test/parse/doctype.test.js
```diff
@@ -44,4 +44,25 @@ describe('doctype', () => {
...
+		<?xml version="1.0">
+		<!-- >'" -->
+		<!DOCTYPE name PUBLIC "identifier" "url" [${internalSubset}]>
+		<name/>
+		`;
+
+		const { parser } = getTestParser();
+		const doc = parser.parseFromString(doctypeString, MIME_TYPE.XML_TEXT);
+
+		expect(doc.doctype).toHaveProperty('internalSubset', internalSubset);
```
karfau, line 66: I formatted the XML snippest in a different way and removed redundant assertions that are already checked in other tests in the same file.
I also slightly changed the assertion to improve the error message in case of a failure.

test/xmltest/__snapshots__/valid.test.js.snap
shunkica: Yes, but beware that my implementation makes absolutely no validation on what is inside the internalSubset. It assumes everything between `[ `and `]` is valid as long as all the comments, brackets and quotes are properly closed.

lib/sax.js
karfau: Yes, everything is fine, I misunderstood your commit message.

karfau (commented):
Overall I'm already quite happe with the PR.
If you find a way to get rid of or simplify the `extractDoctype` function that would be very much appreciated.

I'm also open to pair on it if you want, we just have to find a way to connect.
I have not used Github Codespaces yet, but maybe that's a great option for it.

PS: I'm trying to have a close eye on the PR since I always have to approve the check runs after each commit, since this is your first PR in xmldom. In case I forget/are not fast enough you can @ mention me so I get a notification.

test/xmltest/__snapshots__/valid.test.js.snap
karfau: Yes, that's good enough for me. I just wanted to highlight the other things it fixes.

lib/sax.js
shunkica: @karfau I have pushed the experimental regex parser

shunkica:
> I have not used Github Codespaces yet, but maybe that's a great option for it.

I have to admit I have no idea what that is. I haven't used GitHub much till recently. I'll have to look into it :)

lib/sax.js
```diff
@@ -651,16 +649,22 @@
...
+	// so the values can be serialized as is
+	var regex =
+		/<!DOCTYPE\s+(([:A-Z_a-z\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u02FF\u0370-\u037D\u037F-\u1FFF\u200C-\u200D\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD]|[\uD800-\uDB7F\uDC00-\uDFFF])([:A-Z_a-z\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u02FF\u0370-\u037D\u037F-\u1FFF\u200C-\u200D\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD\-\.0-9\u00B7\u0300-\u036F\u203F-\u2040]|[\uD800-\uDB7F\uDC00-\uDFFF])*)(\s+(SYSTEM)\s+("[^"]*"|'[^']*')|\s+(PUBLIC)\s+("[\r\n\x20\w\-'()+,.\/:=?;!*#@$_%]*"|'[\r\n\x20\w\-()+,.\/:=?;!*#@$_%]*')\s+("[^"]*"|'[^']*'))?\s*(\[((\s*(%(([:A-Z_a-z\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u02FF\u0370-\u037D\u037F-\u1FFF\u200C-\u200D\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD]|[\uD800-\uDB7F\uDC00-\uDFFF])([:A-Z_a-z\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u02FF\u0370-\u037D\u037F-\u1FFF\u200C-\u200D\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF\uFDF0-\uFFFD\-\.0-9\u00B7\u0300-\u036F\u203F-\u2040]|[\uD800-\uDB7F\uDC00-\uDFFF])*);|<!ELEMENT ([^>]*)>|<!ATTLIST ([^>]*)>|<!ENTITY ([^>]*)>|<!NOTATION ([^>]*)>|<\?([^>]*)\?>|<!--([^>]*)-->)\s*)*)])?\s*>/;
```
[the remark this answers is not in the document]
karfau: @shunkica we already have the `NameStart` and `NameChar` in the `conventions` module with an example of composing a regular expression from the source of another one.
Would be nice to use the same approach here to make it more readable.

I'm not sure if that is what the CodeScanning complains about though.

karfau:
I will have a look at the failing tests and the changes you did within the next hours

lib/sax.js
```diff
@@ -651,16 +649,22 @@ ElementAttributes.prototype = {
...
+	var match = regex.exec(source.slice(start));
+	if (!match) {
+		throw new Error('Parse Error: malformed document type');
```
karfau, line 659: inside the sax parser all error should be reported using the `errorHandler`, in this case I would say it is at least an `error` if not a `fatalError`. I need to look up which one it is, since there is a definition for it. I assume this would be an issue of "not well-formed".
In case of the `fatalError` it would would prevent any further processing.

lib/sax.js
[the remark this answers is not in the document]
shunkica: I will look into that.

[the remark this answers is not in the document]
karfau: I know how impressive it is to have such a mighty regular expression, but the amount of time I already invested trying to understand it, tells me it's to complex to maintain in a reasonable way.
I would suggest to split it into multiple pieces that are wrapped by some conditionals, to make it possible to understand what is happening.
e.g. on for getting publicId/systemId, another one for the rest. or if possible split it even further.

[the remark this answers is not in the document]
shunkica: > I know how impressive it is to have such a mighty regular expression, but the amount of time I already invested trying to understand it, tells me it's to complex to maintain in a reasonable way. I would suggest to split it into multiple pieces that are wrapped by some conditionals, to make it possible to understand what is happening. e.g. on for getting publicId/systemId, another one for the rest. or if possible split it even further.

My first approach was tokenizing it but that seemed even worse. It was just lines and lines of if then conditions and after a while I got lost.

I understand that this sausage looks bad, but if it is up to spec I don't see it as something that should necessarily be maintained. Also, tools like [this](https://regex101.com/r/oDPk1V/2) which help visualize the regex make it not look as bad. Its mostly just all the Unicode sequences that are ugly.

If you feel strongly about this I will try something else.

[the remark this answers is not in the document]
shunkica: @karfau Would you consider adding [xregexp](https://github.com/slevithan/xregexp) as a dependancy?
It allows for atomic groups which would solve the backtracking issue reported by CodeQL.
It also allows some other stuff like named capture groups which make the regexp more readable.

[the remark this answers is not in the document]
karfau: So far we do support ES5, which doesn't come with named capture groups.
And we do not use any dependency at runtime, so the step of introducing one is a big one.
The library looks very impressive.
From the docs it sounds like it might be possible to use it as a dev dependency to create the regular expression as a build step, but I'm not really sure about that.

And yes I feel strongly about this, because currently I'm the only active/core maintainer. So at least I have to be able to understand those regular expressions. Otherwise should somebody find any performance or security issue with it, I'm lost.

Whether that means it needs to be split or we need to introduce a dependency or a build step, or what ever it takes, that is something we have to figure out.
Are you up to having a call to discuss it in a sync way instead of going back and forth with comments?
If yes, you can find my mail on my profile to suggest a tool for connecting of your choice.

karfau:
The issue in the regexp has the same message but seems to have changed, since the code scanning marked the previous comment as resolved but created a new one. But the reason for that might also be that's it in a different line now ...

I updated the tests to no longer expect an invalid doctype to work.
I also accepted all the changes to the not well-formed tests, there we expect stuff that should be reported.
But there are still some tests failing in the `tests/xmltest/valid.test.js` which seem to now report a malformed doctype.
Those need to be supported. The XML sources for those tests come from the `xmltest` package devDependency. you can find the zip file containing the sources in the `node_modules`, unpack it and look filenames mentioned in the failing test cases in the folder `xmltest/valid/sa`.

I'm thinking to create a new module containing all needed grammar rules as regular expressions and some reliable mechanism to compose them, so xmldom can consistently rely on applying the rules from the grammar and there is a single place that the grammar is also documented in.
What do you think about that?

karfau:
I did some work on that idea, currently doing some testing.

The main observation/learning already is: the full spec for all doctype grammar can not be done with regular expressions. But is should be close enough to match a string against it to only accept doctypes that most likely are well formed, which hast to be checked step be step afterwards, which can partly be done with regular expressions, but not for all parts.

Should I push those things to your branch or create a separate PR for it once it is ready? (It's quite a big file)

shunkica:
> I did some work on that idea, currently doing some testing.
> 
> The main observation/learning already is: the full spec for all doctype grammar can not be done with regular expressions. But is should be close enough to match a string against it to only accept doctypes that most likely are well formed, which hast to be checked step be step afterwards, which can partly be done with regular expressions, but not for all parts.
> 
> Should I push those things to your branch or create a separate PR for it once it is ready? (It's quite a big file)

I think pushing it here makes more sense. Didn't have any time to work on this but I'll look at your changes tomorrow.

karfau:
I pushed my work.
I'm quite confident about it, but the tests are of course just a start.

And as I said, it's good for making sure the doctype fullfills most WF criteria, but it's not made for getting the values out of it, and I would prefer we do the "extraction part" not from one big RegExp, but either with proper parsing or by building smaller Regular expressions with groups from the pieces we already have.

I will not be able to invest as much time during the week as I did this weekend, I hope it still provides a good start.

test/grammar/doctypedecl.test.js
```diff
@@ -0,0 +1,89 @@
...
+		`<!DOCTYPE name [
+		 <!ELEMENT br EMPTY>
+		 <!ELEMENT p (#PCDATA|emph)* >
+		 <!ELEMENT container ANY>
+		 ]>`,
+		/*
+		TODO: This example comes from the documentation,
+		      but elementdecl only allows `Name` after the first whitespace which doesn't allow % or ;.
+		      Does this mean we need to do entity replacement in elements?
+		 <!ELEMENT %name.para; %content.para; >
```
karfau, line 61: Not sure what do to with this example, even though it comes from the spec, the Grammar looks like this shouldn't be allowed.

lib/grammar.js
```diff
@@ -0,0 +1,320 @@
...
+// [46] contentspec ::= 'EMPTY' | 'ANY' | Mixed | children
+var contentspec = regg('EMPTY', '|', 'ANY', '|', Mixed, '|', children);
+
+// https://www.w3.org/TR/xml11/#NT-elementdecl
+// [45] elementdecl ::= '<!ELEMENT' S Name S contentspec S? '>'
+// https://www.w3.org/TR/xml-names/#NT-elementdecl
+// [17] elementdecl ::= '<!ELEMENT' S QName S contentspec S? '>'
+// because of https://www.w3.org/TR/xml11/#NT-PEReference
+// they also need to support PEReference when checked before replacement of entities declared in DTD (which xmldom is not supporting)
+var elementdecl = reg('<!ELEMENT', S, regg(QName, '|', PEReference), S, regg(contentspec, '|', PEReference), S_OPT, '>');
```
karfau, line 237: @shunkica this is where I documented my understanding why this might be valid (if the parameter entity references have been declared).

shunkica: I saw that, but I still do not understand how a PEReference could appear inside the ELEMENT declaration.

From my understanding a PE reference can appear directly in the internal subset, eg: `<!DOCTYPE x [ %foo; ]>` or inside an entity value, eg: `<!ENTITY foo '%bar;'>`

shunkica:
@karfau I finally found some time to work on this. Let me know what you think.

karfau:
@shunkica we are getting very close, just 5 test failing for XML that should be valid but is currently rejected.

Meanwhile I'm considering to write those regular expressions in typescript and transpile them for bundling, to make it less cumbersome. But this can also be done after this PR works as expected.

shunkica:
> @shunkica we are getting very close, just 5 test failing for XML that should be valid but is currently rejected.

One of those is due to `ENTITIES` missing from `AttType`

> Meanwhile I'm considering to write those regular expressions in typescript and transpile them for bundling, to make it less cumbersome.

I don't understand what exactly you mean by that. Make what less cumbersome?

shunkica:
@karfau where do the valid samples come from? Is this really valid? <!ENTITY e PUBLIC \";!*#@$_%\" \"100.xml\">

karfau:
> @karfau where do the valid samples come from? Is this really valid? `<!ENTITY e PUBLIC \";!*#@$_%\" \"100.xml\">`

They come from this npm package:
https://github.com/karfau/xmltest
So far I whenever I checked a sample it was correct. I will check.

shunkica:
> > @karfau where do the valid samples come from? Is this really valid? `<!ENTITY e PUBLIC \";!*#@$_%\" \"100.xml\">`

I am really combing hard through the spec but I can't figure out how this could be valid.

> > @karfau where do the valid samples come from? Is this really valid? `<!ENTITY e PUBLIC \";!*#@$_%\" \"100.xml\">`
> 
> They come from this npm package: https://github.com/karfau/xmltest So far I whenever I checked a sample it was correct. I will check.

I have downloaded the original files which your package refers to, and there this sample looks like this:

```
<!DOCTYPE doc [
<!ENTITY e PUBLIC ";!*#@$_%" "100.xml">
<!ELEMENT doc (#PCDATA)>
]>
<doc></doc>
```

karfau:
@shunkica I updated my comment above.
it's valid and fixed, 4 left :)

shunkica:
> @shunkica I updated my comment above. it's valid and fixed, 4 left :)

All good. I misinterptered the output of the test. I thought that the quotes are literally escaped (`\"`)in the actual xml

lib/grammar.js
```diff
@@ -0,0 +1,328 @@
...
+
+// [58] NotationType ::= 'NOTATION' S '(' S? Name (S? '|' S? Name)* S? ')'
+//     [VC: Notation Attributes] [VC: One Notation Per Element Type] [VC: No Notation on Empty Element] [VC: No Duplicate Tokens]
+var NotationType = reg('NOTATION', S, /\(/, S_OPT, Name, regg(S_OPT, '|', S_OPT, Name), '*', S_OPT, /\)/);
```
shunkica, line 241:
```suggestion
var NotationType = reg('NOTATION', S, /\(/, S_OPT, Name, regg(S_OPT, /\|/, S_OPT, Name), '*', S_OPT, /\)/);
```

The pipe should be a literal pipe.

lib/grammar.js
```diff
@@ -0,0 +1,328 @@
...
+// [55] StringType ::= 'CDATA'
+// [56] TokenizedType ::= 'ID' [VC: ID] [VC: One ID per Element Type] [VC: ID Attribute Default]
+//   | 'IDREF' [VC: IDREF]
+// | 'IDREFS' [VC: IDREF]
+// | 'ENTITY' [VC: Entity Name]
+// | 'ENTITIES' [VC: Entity Name]
+// | 'NMTOKEN' [VC: Name Token]
+// | 'NMTOKENS' [VC: Name Token]
+// [54] AttType ::= StringType | TokenizedType | EnumeratedType
+var AttType = regg(/CDATA|ID|IDREF|IDREFS|ENTITY|NMTOKEN|NMTOKENS/, '|', EnumeratedType);
```
shunkica, line 257:
```suggestion
var AttType = regg(/CDATA|ID|IDREF|IDREFS|ENTITY|ENTITIES|NMTOKEN|NMTOKENS/, '|', EnumeratedType);
```

`ENTITIES `missing

lib/grammar.js
```diff
@@ -0,0 +1,328 @@
...
+// https://www.w3.org/TR/xml/#document
+// [1] document ::= prolog element Misc*
+// https://www.w3.org/TR/xml11/#NT-document
+// [1] document    ::= ( prolog element Misc* ) - ( Char* RestrictedChar Char* )
+
+// https://www.w3.org/TR/xml/#char32
+// https://www.w3.org/TR/xml11/#NT-Char
+// any Unicode character, excluding the surrogate blocks, FFFE, and FFFF.
+// [2] Char ::= #x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]
+var Char = /[\x09\x0A\x0D\x20-\uD7FF\uE000-\uFFFD]/; // | [#x10000-#x10FFFF]
```
shunkica, line 91:
```suggestion
var Char = /[-\u0001-\u002D\u002E-\uD7FF\uE000-\uFFFD]/; // | [#x10000-#x10FFFF]
```

Fix for `chars_without(Char,'-');`
We should find a way to check for the surrogate pair as well

lib/grammar.js
```diff
@@ -0,0 +1,328 @@
...
+	if (!search) {
+		throw new Error(search + ' is falsy');
+	}
+	if (regexp.source.indexOf(search) === -1) {
+		throw new Error('"' + search + '" is not is /' + regexp.source + '/');
+	}
+	var result = new RegExp(regexp.source.replace(search, ''));
+	result.chars = chars(result);
+	return result;
+}
```
shunkica, line 34: This function should be marked as unsafe, or add some warnings.
For example, if you remove `-`, it will remove it from the first range it encounters, if the literal `-` is not present.

shunkica (commented):
@karfau This should resolve all snapshot issues.

karfau:
> We should find a way to check for the surrogate pair as well

I have not looked into that so far. As long as we stick to ES5 compatibility I assume we would need to leave RegExp land, right?
But if you have any idea, your are welcome to add it.

lib/grammar.js
karfau: I already have it in my local typescript version (only accepting `-` if it is the first char in the character class), will add that.

shunkica: > I already have it in my local typescript version (only accepting `-` if it is the first char in the character class), will add that.

That is just one of the issues.
For example:
```
var test = /[0-9]/
var test_without = chars_without(test, '0')
test_without.source == [-9]
```

shunkica:
> > We should find a way to check for the surrogate pair as well
> 
> I have not looked into that so far. As long as we stick to ES5 compatibility I assume we would need to leave RegExp land, right? But if you have any idea, your are welcome to add it.

I think this should do it:
`var Char = /[-\u0001-\u002D\u002E-\uD7FF\uE000-\uFFFD]|[\uD800-\uDBFF][\uDC00-\uDFFF]/;`

karfau:
> My example is from xml v1.1, but now I see that changed from v1.0
> Is the xml version something this library differentiates? How do you handle this currently?

Currently I always lean towards v1.1 if there are differences.
There is no distinction inside xmldom right now.
(And I'm not aware that anybody has ever requested to strictly only support xml 1.0)

shunkica:
Do you think we could make the parsed internal subset available to the user as well? (the array of strings)
Or would you prefer they were parsed and then exposed similar to `lxml `in python

eg:

```
    # Access the DTD
    dtd = doc_info.internalDTD

    # Access comments
    comments = dtd.itercomments()
    for comment in comments:
        print("Comment:", comment.text)

    # Access elements
    elements = dtd.iterelements()
    for element in elements:
        print("Element:", element.name)

    # Access notations
    notations = dtd.iternotations()
    for notation in notations:
        print("Notation:", notation.name)

    # Access entities
    entities = dtd.iterentities()
    for entity in entities:
        print("Entity:", entity.name)
```

karfau:
I would prefer to just provide access to the string, like it is defined in the DOM Level 2 spec.

So I think we could actually restore the `intSubset` expression (which I recently commented out since it was unused) and simplify the code in the new function to just check for it to match, instead of going through each section.

karfau:
FYI: because of the performance warnings from the code scanning I decided to "simplify" the `children` production to `/[^>]/+`.
Which is also why more "not well-formed" examples now pass again.

Ideally "not well-formed" samples should not be accepted.
This PR is green, so it is already a great improvement.

I'm going to add more fine grained tests for the grammar parts, to make sure we catch breaking changes earlier and with better error messages, and not only when parsing whole DTDs.

karfau:
> I think this should do it:
> `var Char = /[-\u0001-\u002D\u002E-\uD7FF\uE000-\uFFFD]|[\uD800-\uDBFF][\uDC00-\uDFFF]/;`

The issue with that is that it makes quite some expression much more complicated since we are no longer able to treat `Char` as a simple character class. Since it would than contains `|` we would need to wrap it in a group in some contexts where we cannow just add a quantifier after it.
I would not want to handle that in this PR, now that everything is green.

shunkica:
> So I think we could actually restore the `intSubset` expression (which I recently commented out since it was unused) and simplify the code in the new function to just check for it to match, instead of going through each section.

I respectfully disagree. It's my belief that the current method isn't any slower than utilizing the intSubset expression (it might even be faster, although this is just a hunch since I have not bench-marked it). Moreover, it offers the advantage of enabling the extraction of specific sections of the DTD, and on error giving the exact position of the section which caused the error, a feature that could be particularly beneficial for discerning users. It would also be in line with some other libraries like `libxml`.

karfau:
I was able to include the unicode support if the runtime supports it. If the runtime doesn't support it, the additional unicode ranges are not part of the regular expressions.

I guess this is good enough, since even node v10 supports that. (Not all minor versions, but the latest does.)

I even tested it using `rhino` locally and tweaked it to still support it.

shunkica:
Do you need help with anything?

karfau:
The main thing remaining before landing this is to add more unit tests for the regular expressions.
I started with some simple ones yesterday and will most likely be able to cover some expressions every day.
I added a link to a great stack overflow answer on the topic to tests/grammar/utils.js
If you want to join me on that, it would be very helpful.
I plan to cover Name and QName* today.

(The other topic would be to check if there are some (easy as in low hanging fruits) checks we can add to the new function to reject even more of the not well-formed samples.
E.g `PI_unsafe` and the ones using `children` are not really checking for everything the grammar demands.
But this could also go into a follow up PR.)

shunkica:
> I added a link to a great stack overflow answer on the topic to tests/grammar/utils.js

I can't find that.

> (The other topic would be to check if there are some (easy as in low hanging fruits) checks we can add to the new function to reject even more of the not well-formed samples.
> E.g PI_unsafe and the ones using children are not really checking for everything the grammar demands.

The only thing I can do regarding that is write a custom praser.
If I update the regex I need to wait for you to run the checks before I see if it is good or not.
How confident are you about the CodeQL tool, and its ability to analyze regex? Is it possible that those warnings it gave are false positives?

shunkica:
The last commit is just a proposal.
I think this way of parsing the entire doc could potentially resolve a lot of the currently passing not wf.
Only the atomic parts of the XML DOM regex are used (eg. Name, Char, etc..) to validate the contents, and the rest is done in the typical parser fashion.

This parsePI function could resolve https://github.com/xmldom/xmldom/issues/497 if slightly modified and used instead of the current `processingInstruction` function.

karfau:
Sorry, wasn't able to get to anything yesterday and not sure if I have capacity for it today. But I will get back to it this week.

karfau:
> If I update the regex I need to wait for you to run the checks to see if it passes the CodeQL test.

1. It is (most likely) not regarding updating the regular expressions, they can not do more than they already do now. Some grammar rules need to be checked in a different way.
2. Running the test test/grammar/regexp.test.js takes care of recreating test/grammar/regexp.js in case the file is removed (or you temporarilydisable the condition in that test.

karfau:
> > I added a link to a great stack overflow answer on the topic to test/grammar/utils.js
> 
> I can't find that.

There was a commit I forgot to push here is the direct link: https://stackoverflow.com/a/488640/684134

I also updated the snapshots

currently there is no error, but I didn't have time yet to look at your proposal

lib/sax.js
```diff
@@ -576,18 +576,19 @@ function parseDCC(source, start, domBuilder, errorHandler) {
 }
 
 function parseInstruction(source, start, domBuilder) {
-	var end = source.indexOf('?>', start);
```
karfau, line 595: I think this approach is better because `Char*` is greedy, so it would pick as many things as possible, so if there are multiple processing instructions we would only have a singel match. At least this is my current understanding.

And do we also need to prefix `PI_match` with `^` to say that we expect the string to start with it?

lib/sax.js
```diff
@@ -576,18 +576,19 @@ function parseDCC(source, start, domBuilder, errorHandler) {
...
-		var match = source.substring(start, end).match(/^<\?(\S*)\s*([\s\S]*?)\s*$/);
-		if (match) {
-			domBuilder.processingInstruction(match[1], match[2]);
-			return end + 2;
-		} else {
-			//error
-			return -1;
+	var match = source.substring(start).match(grammar.PI_match);
+	if (!match) {
+		throw new ParseError('Invalid processing instruction starting at position ' + start);
```
karfau, line 581: the errors thrown in this method should also be using the `errorHandler.fatalError` instead of throwing themselves.

lib/sax.js
shunkica: I already fixed that with Char*?

`var PI_match = reg(/^<\?/, '(', Name, ')', regg(S, '(', Char, '*?)'), '?', /\?>/);`

karfau: I will double check what ES version is required for this to work, great catch.

lib/sax.js
shunkica: Done. I thought ParseError was expected here since all the code is in a try catch block which catches and rethrows ParseError.

lib/sax.js
shunkica: > And do we also need to prefix `PI_match` with `^` to say that we expect the string to start with it?

Adding `^` here increases performance (for example if regex is `/^<\?/` and input is `<!.....` the regex can return after checking just 2 chars, otherwise it would scan the entire source to find the first <?).
Then we see it is also crucial, because the parser always expects the found match to start at the beginning of the string.
Imagine there are two PI, the first one is invalid, the second one valid, without `^` we would match the second PI, breaking the parser.

karfau:
@shunkica I invited you to the repository, just with "read access" for now, oping that I no longer need to approve the action runs each time, let's see if that works

shunkica:
> @shunkica I invited you to the repository, just with "read access" for now, oping that I no longer need to approve the action runs each time, let's see if that works

Thanks.
Can I ask what the reasoning is for not having those run automatically? Does GitHub charge for that?

karfau:
One of the current errors is related to the fact that there is an expectation for the snippet `ALL_METHODS` in `test/error/xml-reader-dom-handler-errors.test.js` triggers all methods on DOMHandler, including `processingInstruction`, which it currently doesn't.

That is because snippet is detected as not being WF, so it's not even called.
Please find the minimal possible process instruction that would trigger the function and change ALL_METHODS accordingly.

I'm almost certain that this will also solve the other tests in the same file.

Please feel free to push any required changes to the not-wf snapshots for the tests to pass.
You can easily check the local diff after updating them, to see that they all change as expected.

karfau:
Made some progress with the unit tests for the regular expressions
Also splitted the exports in lib/grammar.js into tested and untested ones and sorted all expressions in test/grammar/regexp.js by complexity, so it's easy to decide which one to pick next.

shunkica:
`toErrorSnapshot` is not cross-platform compatible. It doesn't work on windows, because the file path in the stack trace is different.

karfau:
> `toErrorSnapshot` is not cross-platform compatible. It doesn't work on windows, because the file path in the stack trace is different.

Oh, I was not aware. I think we should be able to unify them, as in transforming windows `\` into `/`, so they are the same on all OS.

shunkica:
> > `toErrorSnapshot` is not cross-platform compatible. It doesn't work on windows, because the file path in the stack trace is different.
> 
> Oh, I was not aware. I think we should be able to unify them, as in transforming windows `\` into `/`, so they are the same on all OS.

Here is a stack example on windows:

```
Error: error: entity not found:&e;
    at DOMHandler.<anonymous> (C:\Users\shunkica\Developer\WebStormProjects\xmldom\test\error\reported-levels.test.js:46:23)
    at C:\Users\shunkica\Developer\WebStormProjects\xmldom\node_modules\.pnpm\jest-mock@27.5.1\node_modules\jest-mock\build\index.js:449:39
    at DOMHandler.<anonymous> (C:\Users\shunkica\Developer\WebStormProjects\xmldom\node_modules\.pnpm\jest-mock@27.5.1\node_modules\jest-mock\build\index.js:457:13)
    at DOMHandler.onError (C:\Users\shunkica\Developer\WebStormProjects\xmldom\node_modules\.pnpm\jest-mock@27.5.1\node_modules\jest-mock\build\index.js:114:19)
    at DOMHandler.reportError (C:\Users\shunkica\Developer\WebStormProjects\xmldom\lib\dom-parser.js:446:10)
    at DOMHandler.error (C:\Users\shunkica\Developer\WebStormProjects\xmldom\lib\dom-parser.js:461:8)
    at replace (C:\Users\shunkica\Developer\WebStormProjects\xmldom\lib\sax.js:61:17)
    at String.replace (<anonymous>)
    at appendText (C:\Users\shunkica\Developer\WebStormProjects\xmldom\lib\sax.js:69:42)
    at parse (C:\Users\shunkica\Developer\WebStormProjects\xmldom\lib\sax.js:106:5)
```

karfau:
"Only" two more to be tested:
```
// not unit tested
exports.AttlistDecl = AttlistDecl;
exports.EntityDecl = EntityDecl;
```
I think I should be able to pull off `AttlistDecl` in the next days.
Are you able to look at the tests I wrote so far, try to understand the general approach and contribute to the tests for `EnityDecl`?
This is the most complex one that is exported/used.
It's not important that you finish that work, but figuring out the basics would be a really great help.

Once that is done I would love to do a full review of the PR before finally landing it.

shunkica:
I can't promise anything. It seems like a real time intensive task. I guess could extract the valid and invalid examples from xmltest which I think covers all of these well enough already.

karfau:
OK, then I'll handle it.
We need more fine grained tests to make sure the reg exp is correct, and yes, since we have xmltest samples we don't need complete coverage here or even extract all the samples, but they are a bit to "blackbox"/"many thing could be wrong"-ish for relying only on them.

shunkica:
> the errors thrown in this method should also be using the `errorHandler.fatalError` instead of throwing themselves.

Can you explain something to me?
I think if there is a fatal error, eg. the XML is invalid (not well formed) the parser should exit immediately (throw an error).
What is the point of continuing after encountering the first fatal error?

karfau:
> > the errors thrown in this method should also be using the `errorHandler.fatalError` instead of throwing themselves.
> 
> Can you explain something to me?
> I think if there is a fatal error, eg. the XML is invalid (not well formed) the parser should exit immediately (throw an error).
> What is the point of continuing after encountering the first fatal error?

There is no point.
You can check the implementation of the fatalError, it always tgrows an error and there is no way to prevent this.

The whole point of using tge errorHandler in general is that the code calling the DOMHandler has the chance to be informed before the error is thrown.
This allows that code to decide to also reject XML that reports error or warning levels.
Or to have custom logging/reporting.
This whole idea is part of the SAX specs.

(There are only a few cases in which the code of the sax parser is still throwing directly, but I was not able to trigger those errors, so there is no test for those cases yet.)

shunkica:
@karfau  how do you feel about merging the parse `docTypeDecl` function to the parseDCC function?

lib/sax.js
```diff
@@ -538,86 +538,228 @@ function _copy(source, target) {
...
+			if (skipBlanks() < 1) {
+				return errorHandler.fatalError('No blanks between DOCTYPE and document name');
+			}
+
+			// Parse the DOCTYPE name
+			doctype.name = getMatch(grammar.Name);
+			if (!doctype.name) return errorHandler.fatalError('DOCTYPE without name');
+			skipBlanks();
+
+			// Check for External ID and SystemId
```
karfau, line 644: What do you think about trying to match ^ExternalID once before trying to go through all the steps, it would reduce the number of fine grained error messages, but I think that might be ok?

I mean I also see the benefit of this approach, but it is basically another way of reimplementing the ExternalID regular expression, so we should make sure it stays in sync with it. Which is not trivial from a testing perspective.

(If we keep the detailed error messages we should find a way to put this into a method that we can test individually.)

lib/sax.js
```diff
@@ -538,86 +538,228 @@ function _copy(source, target) {
...
+				return errorHandler.fatalError('Expected <!DOCTYPE');
+			}
+			skip(9);
+			if (skipBlanks() < 1) {
+				return errorHandler.fatalError('No blanks between DOCTYPE and document name');
+			}
+
+			// Parse the DOCTYPE name
+			doctype.name = getMatch(grammar.Name);
+			if (!doctype.name) return errorHandler.fatalError('DOCTYPE without name');
```
karfau, line 641: PUBLIC and SYSTEM would match Name, if the name is left out, should we check for name matching those?

lib/sax.js
```diff
@@ -538,86 +538,228 @@ function _copy(source, target) {
...
+				skip(1);
+				var intSubsetStart = index;
+				skipBlanks();
+				while (index < source.length) {
+					var current = null;
+					// Only in external subset
+					// if (char() === '<' && char(1) === '!' && char(2) === '[') {
+					// 	parseConditionalSections();
+					// } else
+					if (char() === '<' && char(1) === '!') {
```
karfau, line 679: If this is a common case the compare method could also have special handling for text with length 2, avoiding the creation of a substring

lib/sax.js
```diff
@@ -538,86 +538,228 @@ function _copy(source, target) {
...
+					skipBlanks();
+					if (char() === ']') {
+						doctype.internalSubset = source.substring(intSubsetStart, index);
+						skip(1);
+						break;
+					}
+				}
+				if (doctype.internalSubset === null) {
+					return errorHandler.fatalError('Error in internal subset: no closing bracket found');
+				}
```
karfau, line 717: I wonder why this is not happening before setting the value some lines earlier. There is only that one case in which it could be null, since it is initialized with undefined, right?

lib/sax.js
```diff
@@ -538,86 +538,228 @@ function _copy(source, target) {
...
+			}
+			skipBlanks();
+			if (char() !== '>') {
+				return errorHandler.fatalError('DOCTYPE not terminated');
 			}
+			skip(1);
+			domBuilder.startDTD(doctype.name, doctype.publicId, doctype.systemId, doctype.internalSubset);
+			domBuilder.endDTD();
```
karfau, line 725: In the intended way of sax parsing the methods that would be called for the pieces inside the DTD need to be called after startDTD and before endDTD, which is why the internalSubset can not be passed to the startDTD method when we want to start calling these other methods which are currently no-ops.

But it is a totally valid spproach for this PR.

lib/sax.js
```diff
@@ -538,86 +538,228 @@ function _copy(source, target) {
...
+			} else if (compare('SYSTEM')) {
+				skip(6);
+				if (skipBlanks() < 1) {
+					return errorHandler.fatalError('No blanks between SYSTEM and SystemId');
+				}
+				doctype.systemId = getMatch(grammar.SystemLiteral);
+				if (!doctype.systemId) return errorHandler.fatalError('SYSTEM without SystemId');
+			}
+
+			// Parse internal subset
```
karfau, line 666: The only drawback i see with merging the parsing of the whole doctype into the single function is the exploding complexity of the method, which makes it harder/more cumbersome to test.

karfau (commented):
Ideally we should have test cases added for each of the new fatal errors reported.

I like most of the recent changes, especially the improved readability by prefixing calls to fatalError with return.
But let's try really hard to not extend the scope of this PR any further. It's already gigantic.

lib/sax.js
```diff
@@ -538,86 +538,228 @@ function _copy(source, target) {
...
+	function skipBlanks() {
+		var blanks = 0;
+		while (index < source.length) {
+			var c = char();
+			if (c !== ' ' && c !== '\n' && c !== '\t' && c !== '\r') {
+				return blanks;
+			}
+			blanks++;
+			skip();
+		}
```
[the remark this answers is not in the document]
shunkica: I have tested that. Doing it this way is several times faster.

This is what the regex version would look like:

```
  var str = source.substring(index);
  var match = str.match(grammar.reg_start(grammar.S_OPT));
  if (match && match[0]) {
    skip(match[0].length)
    if(match[0].length === str.length) {
      return -1;
    } else {
      return match[0].length;
    }
  } else {
    return 0;
  }
```

For example, when there is only 1 space, the non-regex approach is 60 times faster.
Most of the difference is due to the regex having to be compiled every time.

If the grammar regex is replaced with `var match = str.match(/^[ \t\r\n]*/);` this makes the regex much faster, but still not as fast as the initial while loop approach. The pre-compiled regex only becomes faster when there are 20 or so blanks in a row.

lib/sax.js
```diff
@@ -538,86 +538,228 @@ function _copy(source, target) {
...
+				}
+				doctype.systemId = getMatch(grammar.SystemLiteral);
+				if (!doctype.systemId) return errorHandler.fatalError('SYSTEM without SystemId');
+			}
+
+			// Parse internal subset
+			skipBlanks();
 
-				return lastMatch.index + lastMatch[0].length;
+			if (char() === '[') {
```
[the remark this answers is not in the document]
shunkica: The difference is that compare will always compare from the current index, while char can take a look-ahead parameter, eg. char(2)

lib/sax.js
shunkica: I would need to benchmark this, but if we are going to test the source against `ExternalID`, it doesn't really make sense to parse it this way anyway. We would add matching groups  around `SystemLiteral `and `PubidLiteral `to extract the data then.

lib/sax.js
shunkica: WDYM? Name is required. And i think PUBLIC and SYSTEM are valid names.

```
<!DOCTYPE SYSTEM>
<SYSTEM/>
```

lib/sax.js
shunkica: Yes it should check if it is undefined not null, I think in the previous `parseDoctypeDecl` function I initialized it to null instead of undefined.

lib/sax.js
shunkica: Do you think it would be more readable to make a `DCCParser` class, and move `switch (char(2)) { ... }` into a `parse()` function?

IMO ideally the entire sax parser should be a class, so it can keep track of source, string etc, without them having to be passed around.

lib/sax.js
shunkica: This is less expensive than having an additional check. I would consider using the compare function for >3 chars. and this individual `char(n)` approach for anything less.

shunkica:
I think the regex should be compiled (like it is for tests), and the compiled regex should be used in the lib.

shunkica:
```
const g = require("lib/grammar.js");
const gc = require("grammar/regexp.js");

const sources = [
  "  \r\n\t  \r\n\t  \r\n\t;",
  "!",
  " ",
];
const S_OPT = /[ \t\r\n]*/mu;
const iterations = 1000000;

function fun1(source) {
  const match = source.match(S_OPT);
  if (match && match[0]) {
    return match[0].length;
  } else {
    return 0;
  }
}

function fun2(source) {
  const match = source.match(g.S_OPT);
  if (match && match[0]) {
    return match[0].length;
  } else {
    return 0;
  }
}

function fun3(source) {
  const match = source.match(gc.S_OPT);
  if (match && match[0]) {
    return match[0].length;
  } else {
    return 0;
  }
}

function benchmark(func, iterations, source) {
  const start = performance.now();
  for (let i = 0; i < iterations; i++) {
      func(source);
  }
  const end = performance.now();
  return end - start;
}

sources.forEach((source) => {
  console.log("Source: " + JSON.stringify(source));
  console.log("Time for fun1: " + benchmark(fun1, iterations, source) + "ms");
  console.log("Time for fun2: " + benchmark(fun2, iterations, source) + "ms");
  console.log("Time for fun3: " + benchmark(fun3, iterations, source) + "ms");
});
```

Result:

```
Source: "  \r\n\t  \r\n\t  \r\n\t;"
Time for fun1: 43.16990005970001ms
Time for fun2: 106.15310001373291ms
Time for fun3: 43.21379995346069ms
Source: "!"
Time for fun1: 33.25889992713928ms
Time for fun2: 94.93720006942749ms
Time for fun3: 33.42579996585846ms
Source: " "
Time for fun1: 34.58829998970032ms
Time for fun2: 94.39219999313354ms
Time for fun3: 34.32710003852844ms
```

Using the grammar functions is much slower.
I suggest we compile all regex from grammar during build process, and use that instead of the grammar exports. (probably just move `regexp.js` to lib amd add exports?)

lib/sax.js
karfau: interesting, I was coming from the perspective that a doctype couldwrongly be written as 
`<!DOCTYPE SYSTEM "">` or `<!DOCTYPE PUBLIC "" "">` which would not be well formed, and that we might be able to detect that earlier, but you are right.
`<!DOCTYPE SYSTEM SYSTEM "">` would be well formed, so it doesn't make sense.

lib/sax.js
karfau: Im not proposing to remove the `char` function just that doing a loop in `compare` aka `substringStartsWith` like we do on `skipBlanks` instead of doing the substring and === might make the code easier to understand and maintain and faster. And if we know that there is a limit of e.g. 20 where the loop is slower we could decide based on the length of the string. But I'm not sure we ever pass such long strings to that function, so maybe that check is not needed.
I guess we should also benchmark that.

lib/sax.js
karfau: Yes, I think I would prefer that, just because of the amount of code and the redundancy between that code and the regular expression and all those test cases for the new error cases.

lib/sax.js
karfau: https://github.com/xmldom/xmldom/blob/acf923c9d0853a9fb68e11d4e6ad7c45ad957c59/lib/dom-parser.js#L525-L532

But as I said, not relevant for this PR, so I mark this as resolved

karfau:
> Using the grammar functions is much slower.

I don't really understand why that is the case. they are just using the RegExp constructor when the module is required, so why should it by any faster or slower then the literal regular expression?
is that a general issue of using RegExp constructor vs RegExp literal?

I like that we have benchmarks for things like that, but I tried to do it with a build step and since we don't have any build step at all yet, it introduces an additional level of complexity into how to run tests and what to check before doing a commit/release which I'm not willing to add to this PR.

In the long run it's very likely that we will have a build step, but introducing a "partial build step for this is not my preferred choice, after having tried to do that when we started working on this PR.

lib/sax.js
karfau: the way classes are created in the code base in the ES5 way were there is no class keyword, I would prefer to switch to classes and especially not in this PR.

But we could put this whole "index" and those helper functions into a function that returns an object containing all the functions and a `getIndex`, so it can be used in mutliple places.
It's very close to a class, but I don't see the benefit of the whole `prototype` and `new` stuff.

Alternatively there is the functional approach: Every of the functions that currently modify the index instead receive it as an argument and return the new index. (match could no longer contain skip since it needs to return the match).
I think that would be my preferred solution.

shunkica:
> > Using the grammar functions is much slower.
> 
> I don't really understand why that is the case. they are just using the RegExp constructor when the module is required, so why should it by any faster or slower then the literal regular expression? is that a general issue of using RegExp constructor vs RegExp literal?

I did some more tests and determined the `chars()` function is the culprit for the slowness, it has nothing to do with the regexp being compiled or not

shunkica:
@karfau 
After some additional tests, I am confident in my analysis that assigning a property to the RegExp causes it to be implicitly converted into RegExp on use.

lib/sax.js
shunkica: To be honest I have intentionally refactored this into this pseudo-class style, so it does not require passing the xml and position into every function.
I would very much prefer to leave it like this.
In the event of migrating to ES6 or TS, this code will be easier to port IMO.

lib/sax.js
shunkica: I have benchmarked your proposal:

```
function compare(text, index) {
  return source.substring(index, index + text.length) === text;
}

function compareNew(text, index) {
  for (let i = 0; i < text.length; i++) {
    if (source.charAt(index + i) !== text[i]) {
      return false;
    }
  }
  return true;
}
```

```
Original compare - size 1: 7.834ms
New compare - size 1: 7.179ms

Original compare - size 2: 16.994ms
New compare - size 2: 5.27ms

Original compare - size 3: 9.294ms
New compare - size 3: 7.11ms

Original compare - size 4: 9.846ms
New compare - size 4: 9.396ms
```

From there on original compare starts gaining significant advantage.

It looks like this approach would only be noticeably faster for a string of length 2 and marginally faster for string of length 3.

Given this, I would prefer to leave the compare, (or now `substringStartsWith`) as is.

karfau:
Great finding, might be worth an article.
I will change the code to either not store the chars (create them on the fly each time) or store them as a separate var.

karfau:
The regular expressions no longer have a `chars` property.
THe tests for the EntityDecl are good enough as a first step.

I managed to split the methods again, so they have a managable/testable amount of complexity.
Let me know if the approach I picked works for you. It's not a class, but I would say it's close enough.

The rino example is currently failing, so I have to fix that.

The only other thing missing is to add some unit tests for those new methods and for the new error messages, before the PR can be landed.

karfau (approved):

karfau:
Rhino doesn't seem to support named groups, which we are now using to get understandable results for ExternalIDs...

But I will risk having to change the regular expression and code to make it compatible in a later PR, once it's merged and published, just to see if anybody complains.

karfau:
From my perspective I would land this. Let me know whether you agree or disagree :)

shunkica:
@karfau I agree. Let's proceed with merging, and we'll tackle any issues that arise in subsequent updates.
Thanks for the hard work.

karfau:
landed, I will be able to make a new beta release on the `next` tag in the upcoming days.

karfau:
released as 0.9.0-beta.9 under the `next` version tag

PR #511, opened 2023-07-14
docs: enhance jsdoc comments

shunkica:
This PR aims to add comprehensive JSDoc annotations to dom.js, based on XML DOM spec. The main goal of this enhancement is to improve readability, maintainability, and to provide a more in-depth understanding of the code.

The functionality of the library is unaffected.


karfau:
@shunkica Can you help me to understand why you are adding jsdoc for interfaces that are not implemented by xmldom?

From my understanding just adding the jsdoc without implementing those interfaces only increases the bundle size of the package without providing any value fro the consumers.

If you are planning to use them in some way, I would prefer, if possible, having small PRs that add usage and jsdoc one at a time.

shunkica:
> If you are planning to use them in some way, I would prefer, if possible, having small PRs that add usage and jsdoc one at a time.

Ok. I will keep this PR strictly related to adding jsdoc for functionality which already exists in the lib.

shunkica:
@karfau any idea why the @returns description is inline, and then follows this ugly indentation?

```
/**
 * Adds element as a key to current if it is not already present.
 *
 * @param {Record<string, boolean | undefined>} current
 * The current record object to which the element will be added as a key.
 * The object's keys are string types and values are either boolean or undefined.
 * @param {string} element
 * The string to be added as a key to the current record.
 * @returns {Record<string, boolean | undefined>} The updated record object after the addition
 *                                                of the new element.
 */
```

karfau:
@shunkica I guess this is rather a question for @homer0?

I know that the `@todo` case is also true for me

homer0:
I don't understand the question about the `@returns`; if columns are disabled, and is still trying to push it like that, it's probably a bug.

Regarding the `@todo`, that doesn't sound too complicated, I could take a look over the weekend and see if I can come up with a fix for it.

Something you can try and that it would be really helpful for me to try and understand scenarios like this, is to create a fixture, so I can check it with the functional tests: [validating bugs](https://github.com/homer0/prettier-plugin-jsdoc#-validating-bugs).

shunkica:
I have created a fixture and opened an issue here: https://github.com/homer0/prettier-plugin-jsdoc/issues/3

.gitignore
```diff
@@ -11,3 +11,5 @@ coverage
 # can be used to install node with the correct version when entering the directory
 # https://github.com/direnv/direnv/wiki/Node
 .envrc
+
+/.idea
```
karfau, line 15: Minor: ideally this belongs into your global gitignore file for your user somewhere in your home directory, since it is related to your IDE/local setup

shunkica:
I commit this in my private repos. Will remove it once its ready.

shunkica:
You should go ahead with the release. I am stretched too thin at the moment.

karfau:
What do you think about landing this PR now and you can file the next one once you have time for it?

shunkica:
Did you review all the changes? Are you ok with this?

karfau:
well, currently this PR does to many things at once.

I thought it would only add jsdoc, but that wouldn't explain failing tests.

I would go ahead and revert all changes not related to jsdoc and would be able to land the PR this way. Would that work for you?
(Especially cleaning up all those note type and error code variables should be at least one separate PR. changing the prototype chain of the DOMException should also be an extra PR, and might be considered a breaking one.)

One aspect I'm not sure about is wether it really makes sense to use the type alias `DOMString` for `string`. Since the library doesn't have any extra handling for the things mentioned in https://infra.spec.whatwg.org/#string or https://webidl.spec.whatwg.org/#idl-DOMString I find it more confusing then helpful to use a type alias.
So I would also have to undo the changes related to that, also because it was not applied everywhere consistently.

And last but not least: I don't like the usage of `@todo` tags in doc comments.
I started documenting the difference to the specified behavior as part of the text in a reasoable way and prefer to keep that style.

TODOs should ideally be tracked as github issues, so they can prioritised.
(But of course it would be nice to not go ahead and create an issue for every todo in the codebase. There might already be tickets for some of them, in which case the TODO should be dropped from the code. And it's most likely possible to group some of them into a lower amount of issues.)

Those are quite some changes, but I have them ready.
If it's OK for you I will push them to your branch, and land them.

shunkica:
Sure, go ahead. I trust you will do what is right.

karfau (approved):

karfau:
Thank you for your efforts.

PR #513, opened 2023-07-15
chore: use prettier plugin for jsdoc

shunkica:
This PR includes a commit that adds the prettier-plugin-jsdoc dependency to the project.
The plugin will improve the readability and maintainability of our JSDoc comments and ensure a consistent style across the project.

### DIFF
--- a/lib/conventions.js
+++ b/lib/conventions.js
@@ -5,13 +5,13 @@
  *
  * Works with anything that has a `length` property and index access properties, including NodeList.
  *
- * @template {unknown} T
- * @param {Array<T> | ({length:number, [number]: T})} list
- * @param {function (item: T, index: number, list:Array<T> | ({length:number, [number]: T})):boolean} predicate
- * @param {Partial<Pick<ArrayConstructor['prototype'], 'find'>>?} ac `Array.prototype` by default,
- * 				allows injecting a custom implementation in tests
+ * @param {T[] | { length: number; [number]: T }} list
+ * @param {function (item: T, index: number, list:T[] | ({length:number, [number]: T})):boolean} predicate
+ * @param {Partial<Pick<ArrayConstructor['prototype'], 'find'>>?} ac
+ * `Array.prototype` by default,
+ * allows injecting a custom implementation in tests.
  * @returns {T | undefined}
- *
+ * @template {unknown} T
  * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find
  * @see https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.find
  */
@@ -39,12 +39,13 @@ function find(list, predicate, ac) {
  *
  * Is used to create "enum like" objects.
  *
- * @template T
- * @param {T} object the object to freeze
- * @param {Pick<ObjectConstructor, 'freeze'>} [oc=Object] `Object` by default,
- * 				allows to inject custom object constructor for tests
+ * @param {T} object
+ * The object to freeze.
+ * @param {Pick<ObjectConstructor, 'freeze'>} [oc=Object]
+ * `Object` by default,
+ * allows to inject custom object constructor for tests.
  * @returns {Readonly<T>}
- *
+ * @template T
  * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze
  */
 function freeze(object, oc) {
@@ -60,10 +61,9 @@ function freeze(object, oc) {
  *
  * @param {Object} target
  * @param {Object | null | undefined} source
- *
- * @returns {Object} target
- * @throws TypeError if target is not an object
- *
+ * @returns {Object} Target.
+ * @throws
+ * TypeError if target is not an object.
  * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign
  * @see https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-object.assign
  */
@@ -84,8 +84,8 @@ function assign(target, source) {
  * The presence of a boolean attribute on an element represents the `true` value,
  * and the absence of the attribute represents the `false` value.
  *
- * If the attribute is present, its value must either be the empty string
- * or a value that is an ASCII case-insensitive match for the attribute's canonical name,
+ * If the attribute is present, its value must either be the empty string or a value that is an ASCII case-insensitive match for
+ * the attribute's canonical name,
  * with no leading or trailing whitespace.
  *
  * Note: The values `"true"` and `"false"` are not allowed on boolean attributes.
@@ -126,8 +126,8 @@ var HTML_BOOLEAN_ATTRIBUTES = freeze({
  * This method doesn't check if such attributes are allowed in the context of the current document/parsing.
  *
  * @param {string} name
- * @return {boolean}
- * @see HTML_BOOLEAN_ATTRIBUTES
+ * @returns {boolean}
+ * @see HTML_BOOLEAN_ATTRIBUTES.
  * @see https://html.spec.whatwg.org/#boolean-attributes
  * @see https://html.spec.whatwg.org/#attributes-3
  */
@@ -138,13 +138,28 @@ function isHTMLBooleanAttribute(name) {
 /**
  * Void elements only have a start tag; end tags must not be specified for void elements.
  * These elements should be written as self closing like this: `<area />`.
- * This should not be confused with optional tags that HTML allows to omit the end tag for
- * (like `li`, `tr` and others), which can have content after them,
+ * This should not be confused with optional tags that HTML allows to omit the end tag for (like `li`, `tr` and others), which can
+ * have content after them,
  * so they can not be written as self closing.
  * xmldom does not have any logic for optional end tags cases and will report them as a warning.
  * Content that would go into the unopened element will instead be added as a sibling text node.
  *
- * @type {Readonly<{area: boolean, col: boolean, img: boolean, wbr: boolean, link: boolean, hr: boolean, source: boolean, br: boolean, input: boolean, param: boolean, meta: boolean, embed: boolean, track: boolean, base: boolean}>}
+ * @type {Readonly<{
+ * 	area: boolean;
+ * 	col: boolean;
+ * 	img: boolean;
+ * 	wbr: boolean;
+ * 	link: boolean;
+ * 	hr: boolean;
+ * 	source: boolean;
+ * 	br: boolean;
+ * 	input: boolean;
+ * 	param: boolean;
+ * 	meta: boolean;
+ * 	embed: boolean;
+ * 	track: boolean;
+ * 	base: boolean;
+ * }>}
  * @see https://html.spec.whatwg.org/#void-elements
  * @see https://html.spec.whatwg.org/#optional-tags
  */
@@ -167,12 +182,11 @@ var HTML_VOID_ELEMENTS = freeze({
 
 /**
  * Check if `tagName` is matching one of the HTML void element names.
- * This method doesn't check if such tags are allowed
- * in the context of the current document/parsing.
+ * This method doesn't check if such tags are allowed in the context of the current document/parsing.
  *
  * @param {string} tagName
- * @return {boolean}
- * @see HTML_VOID_ELEMENTS
+ * @returns {boolean}
+ * @see HTML_VOID_ELEMENTS.
  * @see https://html.spec.whatwg.org/#void-elements
  */
 function isHTMLVoidElement(tagName) {
@@ -183,8 +197,8 @@ function isHTMLVoidElement(tagName) {
  * Tag names that are raw text elements according to HTML spec.
  * The value denotes whether they are escapable or not.
  *
- * @see isHTMLEscapableRawTextElement
- * @see isHTMLRawTextElement
+ * @see IsHTMLEscapableRawTextElement.
+ * @see IsHTMLRawTextElement.
  * @see https://html.spec.whatwg.org/#raw-text-elements
  * @see https://html.spec.whatwg.org/#escapable-raw-text-elements
  */
@@ -197,13 +211,12 @@ var HTML_RAW_TEXT_ELEMENTS = freeze({
 
 /**
  * Check if `tagName` is matching one of the HTML raw text element names.
- * This method doesn't check if such tags are allowed
- * in the context of the current document/parsing.
+ * This method doesn't check if such tags are allowed in the context of the current document/parsing.
  *
  * @param {string} tagName
- * @return {boolean}
- * @see isHTMLEscapableRawTextElement
- * @see HTML_RAW_TEXT_ELEMENTS
+ * @returns {boolean}
+ * @see IsHTMLEscapableRawTextElement.
+ * @see HTML_RAW_TEXT_ELEMENTS.
  * @see https://html.spec.whatwg.org/#raw-text-elements
  * @see https://html.spec.whatwg.org/#escapable-raw-text-elements
  */
@@ -213,13 +226,12 @@ function isHTMLRawTextElement(tagName) {
 }
 /**
  * Check if `tagName` is matching one of the HTML escapable raw text element names.
- * This method doesn't check if such tags are allowed
- * in the context of the current document/parsing.
+ * This method doesn't check if such tags are allowed in the context of the current document/parsing.
  *
  * @param {string} tagName
- * @return {boolean}
- * @see isHTMLRawTextElement
- * @see HTML_RAW_TEXT_ELEMENTS
+ * @returns {boolean}
+ * @see IsHTMLRawTextElement.
+ * @see HTML_RAW_TEXT_ELEMENTS.
  * @see https://html.spec.whatwg.org/#raw-text-elements
  * @see https://html.spec.whatwg.org/#escapable-raw-text-elements
  */
@@ -232,7 +244,6 @@ function isHTMLEscapableRawTextElement(tagName) {
  *
  * @param {string} mimeType
  * @returns {mimeType is 'text/html'}
- *
  * @see https://www.iana.org/assignments/media-types/text/html
  * @see https://en.wikipedia.org/wiki/HTML
  * @see https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString
@@ -242,12 +253,11 @@ function isHTMLMimeType(mimeType) {
 	return mimeType === MIME_TYPE.HTML;
 }
 /**
- * For both the `text/html` and the `application/xhtml+xml` namespace
- * the spec defines that the HTML namespace is provided as the default.
+ * For both the `text/html` and the `application/xhtml+xml` namespace the spec defines that the HTML namespace is provided as the
+ * default.
  *
  * @param {string} mimeType
  * @returns {boolean}
- *
  * @see https://dom.spec.whatwg.org/#dom-document-createelement
  * @see https://dom.spec.whatwg.org/#dom-domimplementation-createdocument
  * @see https://dom.spec.whatwg.org/#dom-domimplementation-createhtmldocument
@@ -261,7 +271,7 @@ function hasDefaultHTMLNamespace(mimeType) {
  *
  * @see https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString#Argument02 MDN
  * @see https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#domparsersupportedtype WHATWG HTML Spec
- * @see DOMParser.prototype.parseFromString
+ * @see DOMParser.prototype.parseFromString.
  */
 var MIME_TYPE = freeze({
 	/**
@@ -312,11 +322,11 @@ var MIME_TYPE = freeze({
 	XML_SVG_IMAGE: 'image/svg+xml',
 });
 /**
- * @typedef {'application/xhtml+xml' | 'application/xml' | 'image/svg+xml' |  'text/html' | 'text/xml'} MimeType
+ * @typedef {'application/xhtml+xml' | 'application/xml' | 'image/svg+xml' | 'text/html' | 'text/xml'} MimeType
  */
 /**
  * @type {MimeType[]}
- * @private
+ * @access private
  */
 var _MIME_TYPES = Object.keys(MIME_TYPE).map(function (key) {
 	return MIME_TYPE[key];
@@ -324,6 +334,7 @@ var _MIME_TYPES = Object.keys(MIME_TYPE).map(function (key) {
 
 /**
  * Only returns true if `mimeType` is one of the allowed values for `DOMParser.parseFromString`.
+ *
  * @param {string} mimeType
  * @returns {mimeType is 'application/xhtml+xml' | 'application/xml' | 'image/svg+xml' |  'text/html' | 'text/xml'}
  */
@@ -358,7 +369,7 @@ var NAMESPACE = freeze({
 	XML: 'http://www.w3.org/XML/1998/namespace',
 
 	/**
-	 * The `xmlns:` namespace
+	 * The `xmlns:` namespace.
 	 *
 	 * @see https://www.w3.org/2000/xmlns/
 	 */
@@ -368,9 +379,10 @@ var NAMESPACE = freeze({
 /**
  * Creates an error that will not be caught by XMLReader aka the SAX parser.
  *
+ * @class
  * @param {string} message
- * @param {any?} locator Optional, can provide details about the location in the source
- * @constructor
+ * @param {any?} locator
+ * Optional, can provide details about the location in the source.
  */
 function ParseError(message, locator) {
 	this.message = message;
--- a/lib/conventions.js
+++ b/lib/conventions.js
@@ -5,12 +5,12 @@
  *
  * Works with anything that has a `length` property and index access properties, including NodeList.
  *
+ * @template {unknown} T
  * @param {T[] | { length: number; [number]: T }} list
  * @param {function (item: T, index: number, list:(T)[] | ({length:number, [number]: T})):boolean} predicate
- * @param {Partial<Pick<ArrayConstructor['prototype'], 'find'>>?} ac
- * `Array.prototype` by default, allows injecting a custom implementation in tests.
+ * @param {Partial<Pick<ArrayConstructor['prototype'], 'find'>>?} ac `Array.prototype` by default, allows injecting a custom
+ *   implementation in tests
  * @returns {T | undefined}
- * @template {unknown} T
  * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find
  * @see https://tc39.es/ecma262/multipage/indexed-collections.html#sec-array.prototype.find
  */
@@ -37,11 +37,11 @@ function find(list, predicate, ac) {
  *
  * Is used to create "enum like" objects.
  *
- * @param {T}                                 object       The object to freeze.
- * @param {Pick<ObjectConstructor, 'freeze'>} [oc=Object]  `Object` by default, allows to inject custom object constructor for
- *                                                         tests. Default is `Object`
- * @returns {Readonly<T>}
  * @template T
+ * @param {T} object The object to freeze
+ * @param {Pick<ObjectConstructor, 'freeze'>} [oc=Object] `Object` by default, allows to inject custom object constructor for
+ *   tests. Default is `Object`
+ * @returns {Readonly<T>}
  * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze
  */
 function freeze(object, oc) {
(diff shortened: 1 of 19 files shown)


karfau (commented):
I'm not sure where those options should go (assume to the top level?), but I would prefer to enable the `tsdoc` and `jsdocPreferCodeFences` options.

karfau:
@shunkica before you invest eny more time into this, I just had a closer look at the plugin from a dependency perspective, and there are some things that make me feel uncomfortable:
- quite some number of reported issues (have you used it with a good/stable devx before?)
- not much maintainer activity in the recent month at first glance
- no compatibility with prettier v3 (I was hoping/planning to land #504 soonish)
- quite some issues detected on socket.dev (see https://socket.dev/npm/package/prettier-plugin-jsdoc/overview/0.4.2) I think that's mostly because all it's dev dependencies are installed as dependencies, which is why it adds 36 dependencies to our already long list of devDependencies, which I would prefer to avoid.

That's not a "NO", but some things to consider.

I think I will file a PR over there that moves develeopment dependencies to the `devDependencies` part of `package.json`, and if this is landed and/or prettier v3 support is fixed (there is already a PR), I'have no issues with adding it.

shunkica:
I have not used this plugin before, but since it is only a dev dependency, has over 40000 weekly downloads, and it did what it is supposed to, I have not given it that much thought.
If you have issues with that plugin, maybe this one would be better as it doesn't have that many dependencies, but it is a much (much) smaller project, and I am not sure if it supports prettier v3 or not. [@homer0/prettier-plugin-jsdoc](https://github.com/homer0/packages/tree/main/packages/public/prettier-plugin-jsdoc)

karfau:
That looks much more reliable and alive, even though there are not as many downloads.
It is also not supporting prettier v3 yet, and v6 dropped node v14 support, but I think prettier v3 also requires node v16

homer0:
@karfau 👋 I just published the support for v3, and I'm planning to drop v16 before end of the year (since it's reaching EOL on October).

shunkica:
[@homer0](https://github.com/homer0) I have added `@homer0/prettier-plugin-jsdoc` `v7.0.0` to the project, but when running format this error occurs a bunch of times.
```

TypeError: Cannot read properties of undefined (reading 'startsWith')
[error]     at printAngular (/home/ivanclab/DevProjects/external/xmldom/node_modules/prettier/index.js:24775:22)
[error]     at printPathNoParens (/home/ivanclab/DevProjects/external/xmldom/node_modules/prettier/index.js:29687:25)
[error]     at Object.genericPrint [as print] (/home/ivanclab/DevProjects/external/xmldom/node_modules/prettier/index.js:29629:23)
[error]     at callPluginPrintFunction (/home/ivanclab/DevProjects/external/xmldom/node_modules/prettier/index.js:8601:26)
[error]     at mainPrintInternal (/home/ivanclab/DevProjects/external/xmldom/node_modules/prettier/index.js:8550:22)
[error]     at mainPrint (/home/ivanclab/DevProjects/external/xmldom/node_modules/prettier/index.js:8537:18)
[error]     at printAstToDoc (/home/ivanclab/DevProjects/external/xmldom/node_modules/prettier/index.js:8529:18)
[error]     at coreFormat (/home/ivanclab/DevProjects/external/xmldom/node_modules/prettier/index.js:8837:20)
[error]     at formatWithCursor2 (/home/ivanclab/DevProjects/external/xmldom/node_modules/prettier/index.js:9021:18)
[error]     at Object.formatWithCursor (/home/ivanclab/DevProjects/external/xmldom/node_modules/prettier/index.js:38183:12)
```

karfau:
You might have to also bump prettier and eslint versions at the same time to make it work with v7.

shunkica:
> You might have to also bump prettier and eslint versions at the same time to make it work with v7.
> 
> Or try to make it work with v6 first

Yes I figured that out. Do you want to use `@homer0/prettier-plugin-jsdoc:6.0.5` and keep prettier version, or use `7.0.0` and bump prettier version?

karfau:
Let's first make it work with the current prettier version, I can take care if the updates afterwards

lib/dom-parser.js
```diff
@@ -49,98 +41,281 @@ function normalizeLineEndings(input) {
...
+ *                                                                                                                                used,
+ *                                                                                                                                will
+ *                                                                                                                                be
+ *                                                                                                                                overridden
+ *                                                                                                                                according
+ *                                                                                                                                to
+ *                                                                                                                                the
+ *                                                                                                                                specification.
+ * @see NormalizeLineEndings.
  */
```
shunkica, line 271: @homer0 any idea what happened here?

homer0:
I’ll take a look tonight (I’m GMT-3), but I tested it with the cli, vscode
ext and my e2e suite and didn’t see anything like that.

Most of the issues I had were about the code not changing, since Prettier
made its api async and I wasn’t awaiting stuff.

I’ll let you know how it goes.

shunkica:
> I’ll take a look tonight (I’m GMT-3), but I tested it with the cli, vscode ext and my e2e suite and didn’t see anything like that. Most of the issues I had were about the code not changing, since Prettier made its api async and I wasn’t awaiting stuff. I’ll let you know how it goes.

Thanks, please note though that this happened with version `6.0.5` I have not tested `7.0.0` since we are still on prettier v2.

homer0:
@shunkica 👋 

I just took a look and this line was the issue:

```ts
/*
 * @property {typeof conventions.assign} [assign=Object.assign || conventions.assign]
 * The method to use instead of `Object.assign` (or if not available `conventions.assign`),
 * which is used to copy values from the options before they are used for parsing.
 */
```

Adding a parenthesis o r changing the default value of `assign` fixes the issue when formatting.

I know, it's not great that something like that would break the whole formatting, but that tends to happen when you start mixing a lot of TS functions with JSDoc, like that OR in the default value.

Some stuff may work, like the `typeof` on the type, and that's mostly magic from the IDE/Editor, but it can break other things, for example: You have the `onError` property on that same object, but if you check the inferred type, you'll see that the property doesn't have a name:

image

So the reference on `onErrorStopParsing` doesn't properly link there. That's because that TS syntax don't work on all context; you need to change the type to this in order to make it properly work:

```ts
/*
 * @property {(level:ErrorLevel, message:string, context: DOMHandler) => void} [onError]
 */
```

image

My recommendation, regardless of the formatting plugin you end up choosing, would be to always check with the IDE/Editor to make sure the inferred type make sense, and to be very careful with the default values, because I believe the editors will only recognize the brackets (making the param/property optional) and ignore the default value, so it will fallback to standard JSDoc, which doesn't accept very complex types.

I hope that helps :)

karfau:
Thx a lot, if adding the default value this way is not reliable we have to do it differently.
@shunkica if you are not able to get it working let me know and I will have a look

shunkica:
There is also some kind of conflict between eslint and prettier now, because they produce different versions of the jsdoc. After running prettier format, eslint complains about a bunch of stuff (spaces and such)

karfau:
Since it is the eslint prettier/prettier rule that is complaining, I will take care of it by separating eslint and prettier.
From my experience in other projects this is what most people do now anyways, also for eslint performance problems when using it to run prettier.

shunkica:
It turns out the column option was the one causing a bit of a hiccup. I've decided to turn it off for now, as it wasn't really meshing with the overall aesthetic. Hope that's alright!

lib/dom-parser.js
```diff
@@ -17,22 +17,16 @@ var ParseError = conventions.ParseError;
...
- * >
- * > To simplify the tasks of applications, the XML processor must behave
- * > as if it normalized all line breaks in external parsed entities (including the document entity)
- * > on input, before parsing, by translating all of the following to a single #xA character:
- * >
- * > 1. the two-character sequence #xD #xA
- * > 2. the two-character sequence #xD #x85
- * > 3. the single character #x85
- * > 4. the single character #x2028
- * > 5. any #xD character that is not immediately followed by #xA or #x85.
```
karfau, line 35: Ouch, why was the block quote dropped from the docstring?

If this is what the plugin does, I think we need to add `@prettierignore` to this doc comment.

lib/dom-parser.js
```diff
@@ -49,98 +43,82 @@ function normalizeLineEndings(input) {
 
 /**
  * @typedef DOMParserOptions
- * @property {typeof conventions.assign} [assign=Object.assign || conventions.assign]
- * The method to use instead of `Object.assign` (or if not available `conventions.assign`),
- * which is used to copy values from the options before they are used for parsing.
+ * @property {typeof assign} assign
+ * The method to use instead of `Object.assign`
+ * (or if not available `conventions.assign`), which is used to copy values from the options before they are used for parsing.
+ * Default is `Object.assign || conventions.assign`
```
karfau, line 49: It's fine to move the description of the default value to the text, but the parameter needs to still be marked as optional.

And completely unrelated to this PR I just realized that the described default is not up to date with the code, so it should be changed to
```suggestion
 * @property {typeof assign} [assign]
 * The method to use instead of `conventions.assign`, which is used to copy values from `options` before they are used for parsing.
 * Default is `conventions.assign`
```

lib/dom-parser.js
```diff
@@ -49,98 +43,82 @@ function normalizeLineEndings(input) {
...
 	/**
-	 * The method to use instead of `Object.assign` (defaults to or `conventions.assign`),
-	 * which is used to copy values from the options before they are used for parsing.
+	 * The method to use instead of `Object.assign` (defaults to or `conventions.assign`), which is used to copy values from the
+	 * options before they are used for parsing.
 	 *
-	 * @type {function (target: object, source: object | null | undefined): object}
+	 * @type {function (target: Object, source: Object | null | undefined): Object}
+	 * @access private
+	 * @see Conventions.assign.
```
karfau, line 95: Let's reduce some redundancy here and do it the same way as above:
```suggestion
	 * @type {typeof assign}
	 * @access private
	 * @see conventions.assign.
```

lib/dom-parser.js
```diff
@@ -49,98 +43,82 @@ function normalizeLineEndings(input) {
...
  * @see https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#dom-parsing-and-serialization
  */
 function DOMParser(options) {
 	options = options || { locator: true };
 
 	/**
-	 * The method to use instead of `Object.assign` (defaults to or `conventions.assign`),
-	 * which is used to copy values from the options before they are used for parsing.
+	 * The method to use instead of `Object.assign` (defaults to or `conventions.assign`), which is used to copy values from the
+	 * options before they are used for parsing.
```
karfau, line 91: Should also be synced with the comment suggestion from above: 
```suggestion
	 * The method to use instead of `conventions.assign`, which is used to copy values from `options` before they are used for parsing.
   * Default is `conventions.assign`
```

lib/dom-parser.js
```diff
@@ -49,98 +43,82 @@ function normalizeLineEndings(input) {
...
-	 * @private
-	 * @see conventions.assign
 	 */
 	this.assign = options.assign || conventions.assign;
 
 	/**
-	 * For internal testing: The class for creating an instance for handling events from the SAX parser.
-	 * __**Warning: By configuring a faulty implementation, the specified behavior can completely be broken.**__
+	 * For internal testing: The class for creating an instance for handling events from the SAX parser. ****Warning: By configuring
+	 * a faulty implementation, the specified behavior can completely be broken.****
```
karfau, line 102: The original intention was to make it bold and italic :)

```suggestion
	 * For internal testing: The class for creating an instance for handling events from the SAX parser. *****Warning: By configuring
	 * a faulty implementation, the specified behavior can completely be broken.*****
```

lib/dom-parser.js
```diff
@@ -49,98 +43,82 @@ function normalizeLineEndings(input) {
...
-	 * Be aware that many `warning`s are considered an error
-	 * that prevents further processing in most implementations.
+	 * Be aware that many `warning`s are considered an error that prevents further processing in most implementations.
 	 *
 	 * @type {function(level:ErrorLevel, message:string, context: DOMHandler):void}
-	 *
-	 * @see onErrorStopParsing
-	 * @see onWarningStopParsing
+	 * @see OnErrorStopParsing.
+	 * @see OnWarningStopParsing.
```
karfau, line 121: @shunkica @homer0 Is camel casing the first letter coming from the plugin?
This would be a no-go for me. Or do we need to keep those references to imported things in single quotes to indicate that the plugin shouldn't tough those?

lib/dom-parser.js
```diff
@@ -248,83 +223,79 @@ DOMParser.prototype.parseFromString = function (source, mimeType) {
...
-	 * to store the current parsing position.
-	 * It is used by DOMHandler to set `columnNumber` and `lineNumber`
-	 * on the DOM nodes.
+	 * The locator is stored as part of setDocumentLocator. It is controlled and mutated by the SAX parser to store the current
+	 * parsing position. It is used by DOMHandler to set `columnNumber` and `lineNumber` on the DOM nodes.
 	 *
 	 * @type {Readonly<Locator> | undefined}
-	 * @readonly (the sax parser currently sometimes set's it)
-	 * @private
+	 * @access private
```
karfau, line 296: I don't see any mention of the `@access` tag in the typescript docs: https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#property-modifiers

Is it possible to turn off the conversion of `@private` to `@access private`?

lib/entities.js
```diff
@@ -2160,7 +2161,8 @@ exports.HTML_ENTITIES = freeze({
 });
 
 /**
- * @deprecated use `HTML_ENTITIES` instead
- * @see HTML_ENTITIES
+ * @see HTML_ENTITIES.
+ * @deprecated
+ * Use `HTML_ENTITIES` instead.
```
karfau, line 2166: I really want `@deprecated` to always be the first tag to be listed if it is provided, so we should configure the tag order.

lib/grammar.js
```diff
@@ -207,20 +211,18 @@ var NCNameChar = chars_without(NameChar, ':');
...
+ *     [7] QName ::= PrefixedName | UnprefixedName
+ *     				  === (NCName ':' NCName) | NCName
+ *     				  === NCName (':' NCName)?
+ *     [8] PrefixedName ::= Prefix ':' LocalPart
+ *     								 === NCName ':' NCName
+ *     [9] UnprefixedName ::= LocalPart
+ *     									 === NCName
+ *     [10] Prefix ::= NCName
+ *     [11] LocalPart ::= NCName
+ */
```
karfau, line 225: I really prefer  to stick to the "code fences" (wrapping those things by triple back ticks) way of doing that, since I really dislike this mixture of spaces and tabs in the same line.

I think there is an option for this.

lib/sax.js
```diff
@@ -415,7 +415,7 @@ function parseElementStartPart(source, start, el, currentNSMap, entityReplacer,
 }
 
 /**
- * @return true if has new namespace define
+ * @returns True if has new namespace define.
```
karfau, line 418: I assume the following makes more sense/is easier to understand?
```suggestion
 * @returns `true` if a new namespace has been defined.
```

karfau (commented):
I didn't went through all of the changes, sicne there are already quite some comments/questions.

Feel free to ignore the "content corrections" if you think they are in the way of getting the formatting right, we can also apply them later.

Thx for investing the time, I hope we can find a way to resolve all the open questions and can come up with a config that provides consistency without getting in our way.

lib/conventions.js
```diff
@@ -5,13 +5,12 @@
...
- * @param {Array<T> | ({length:number, [number]: T})} list
- * @param {function (item: T, index: number, list:Array<T> | ({length:number, [number]: T})):boolean} predicate
- * @param {Partial<Pick<ArrayConstructor['prototype'], 'find'>>?} ac `Array.prototype` by default,
- * 				allows injecting a custom implementation in tests
+ * @param {T[] | { length: number; [number]: T }} list
+ * @param {function (item: T, index: number, list:(T)[] | ({length:number, [number]: T})):boolean} predicate
+ * @param {Partial<Pick<ArrayConstructor['prototype'], 'find'>>?} ac
+ * `Array.prototype` by default, allows injecting a custom implementation in tests.
  * @returns {T | undefined}
- *
```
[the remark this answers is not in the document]
karfau: Or would an empty line between tags be preserved, so we can just "preserve them by adding them back manually once?

lib/conventions.js
```diff
@@ -33,18 +32,17 @@ function find(list, predicate, ac) {
...
  * Is used to create "enum like" objects.
  *
- * @template T
- * @param {T} object the object to freeze
- * @param {Pick<ObjectConstructor, 'freeze'>} [oc=Object] `Object` by default,
- * 				allows to inject custom object constructor for tests
+ * @param {T} object
+ * The object to freeze.
+ * @param {Pick<ObjectConstructor, 'freeze'>} [oc=Object]
+ * `Object` by default, allows to inject custom object constructor for tests. Default is `Object`
```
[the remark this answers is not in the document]
homer0: Is not the plugin :). It doesn't add new text to the comments.

lib/conventions.js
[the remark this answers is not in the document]
shunkica: I don't mind that honestly.
Adding this manually will not fix it, because eslint will complain, and another format will remove the line anyway.

lib/conventions.js
```diff
@@ -33,18 +32,17 @@ function find(list, predicate, ac) {
 }
 
 /**
- * "Shallow freezes" an object to render it immutable.
- * Uses `Object.freeze` if available,
- * otherwise the immutability is only in the type.
+ * "Shallow freezes" an object to render it immutable. Uses `Object.freeze` if available, otherwise the immutability is only in
+ * the type.
```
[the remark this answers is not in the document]
shunkica: There is a `jsdocPrintWidth` option which overrides the `printWidth` option.

lib/conventions.js
```diff
@@ -55,15 +53,13 @@ function freeze(object, oc) {
...
  *
  * @param {Object} target
  * @param {Object | null | undefined} source
- *
- * @returns {Object} target
- * @throws TypeError if target is not an object
- *
+ * @returns {Object} Target.
+ * @throws
+ * TypeError if target is not an object.
```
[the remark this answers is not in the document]
homer0: This happens for two reasons:

1. I believe the columns were disabled, so the format is now `@tag {type}\ndescription`
2. The type is not between curly brackets:

```
@throws free-form description
@throws {<type>}
@throws {<type>} free-form description
```
> [`@throws` doc](https://jsdoc.app/tags-throws.html)

So, this should be `@throws {TypeError} If target is not an object.`

lib/conventions.js
```diff
@@ -122,12 +116,12 @@ var HTML_BOOLEAN_ATTRIBUTES = freeze({
...
- * Check if `name` is matching one of the HTML boolean attribute names.
- * This method doesn't check if such attributes are allowed in the context of the current document/parsing.
+ * Check if `name` is matching one of the HTML boolean attribute names. This method doesn't check if such attributes are allowed
+ * in the context of the current document/parsing.
  *
  * @param {string} name
- * @return {boolean}
- * @see HTML_BOOLEAN_ATTRIBUTES
+ * @returns {boolean}
+ * @see HTML_BOOLEAN_ATTRIBUTES.
```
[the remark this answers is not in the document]
homer0: Like `@throws`, the type is not wrapper in curly brackets, so it's assuming it's a description:

```
@see {@link foo} for further information.
@see {@link http://github.com|GitHub}
```
> [`@see` doc](https://jsdoc.app/tags-see.html)

So, this should be `@see {@link HTML_BOOLEAN_ATTRIBUTES}`

lib/conventions.js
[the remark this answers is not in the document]
shunkica: Honestly, it's quite possible that you're analyzing this a bit too much right now. Just a friendly observation! :)

lib/conventions.js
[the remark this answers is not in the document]
shunkica: I believe it was added by the previous plugin. I could revert all changes and just do the format with this plugin.

lib/conventions.js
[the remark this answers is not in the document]
shunkica: This is something that bothers me as well. It seems there is no way to do this. As soon as columns are disabled, it forces all descriptions to be on newlines.

lib/dom-parser.js
homer0: Yeah, I'd use `@prettierignore` for this; there's a limit of how much magic can the plugin handle and, although using the `> ` as a prefix and running the same logic as for descriptions wouldn't be complicated, I don't want to open the door to "supports markdown", because fully supporting it would be a pain :P.

lib/conventions.js
```diff
@@ -80,16 +76,14 @@ function assign(target, source) {
...
- * If the attribute is present, its value must either be the empty string
- * or a value that is an ASCII case-insensitive match for the attribute's canonical name,
- * with no leading or trailing whitespace.
+ * If the attribute is present, its value must either be the empty string or a value that is an ASCII case-insensitive match for
+ * the attribute's canonical name, with no leading or trailing whitespace.
  *
- * Note: The values `"true"` and `"false"` are not allowed on boolean attributes.
- * To represent a `false` value, the attribute has to be omitted altogether.
+ * Note: The values `"true"` and `"false"` are not allowed on boolean attributes. To represent a `false` value, the attribute has
+ * to be omitted altogether.
```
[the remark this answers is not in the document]
shunkica: This was a byproduct of the other plugin as well. I have reverted all changes made by the other plugin, and formatted again from the scratch with the homer0 plugin.

lib/dom-parser.js
homer0: This is the curly brackets thing I mentioned in the other comment. It thinks that those types are description, and since `jsdocEnsureDescriptionsAreSentences` is enabled by default, it's making the first letter uppercase, and ensuring there's a period at the end.

lib/conventions.js
```diff
@@ -368,9 +369,10 @@ var NAMESPACE = freeze({
 /**
  * Creates an error that will not be caught by XMLReader aka the SAX parser.
  *
+ * @class
  * @param {string} message
- * @param {any?} locator Optional, can provide details about the location in the source
- * @constructor
+ * @param {any | null} locator
```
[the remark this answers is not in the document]
shunkica: These were changes from the old plugin. The new plugin did not change the original:
`@param {any?} locator`

lib/dom-parser.js
homer0: Just in case, the plugin is based on JSDoc, not the JSDoc support on TSDoc.

You can use [`jsdocAllowAccessTag `](https://github.com/homer0/packages/tree/main/packages/public/prettier-plugin-jsdoc#allow-access-tag) to control the access tag.

lib/dom-parser.js
shunkica: Also byproduct of the other plugin.

lib/dom-parser.js
shunkica: Have a look now.

lib/dom-parser.js
shunkica: Also byproduct of the other plugin.

lib/dom-parser.js
shunkica: It is, we need to add `@link` for what you want to achieve.

lib/dom-parser.js
shunkica: There is. [`jsdocEnforceAccessTag`](https://github.com/homer0/packages/blob/main/packages/public/prettier-plugin-jsdoc/README.md#enfore-the-use-of-the-access-tag)

lib/entities.js
shunkica: There is [`jsdocTagsOrder`](https://github.com/homer0/packages/blob/main/packages/public/prettier-plugin-jsdoc/README.md#tags-order)

lib/grammar.js
shunkica: Byproduct of other plugin.

lib/conventions.js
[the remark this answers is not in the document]
homer0: Not sure if there was some manual modification in there, but I just took the code for that from `master`, ran the plugin, and the only change was this:

```diff
- * If the attribute is present, its value must either be the empty string
- * or a value that is an ASCII case-insensitive match for the attribute's canonical name,
- * with no leading or trailing whitespace.
+ * If the attribute is present, its value must either be the empty string or a value that is an ASCII case-insensitive match for
+ * the attribute's canonical name,
+ * with no leading or trailing whitespace.
```

Because the default `printWidth` is `130`, so there was a lot of room in there for more words; but check that it respects the comma:


https://github.com/xmldom/xmldom/assets/6949310/44360328-4428-41fc-95b6-22ee491a2360

(Ignore the error on the actual JS code, that's because I took that snippet and placed in a temp file)
(I ran it with v6)

lib/conventions.js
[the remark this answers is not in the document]
shunkica: Actually it looks like it does respect commas. Also byproduct of other plugin.

lib/conventions.js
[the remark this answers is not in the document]
shunkica: @homer0 
`@throws {<type>} free-form description` does not work either when columns are disabled. It will always put description in new line.

lib/conventions.js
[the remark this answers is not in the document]
homer0: It matches any character that is not `a-zA-Z0-9_` (`\W`):


https://github.com/xmldom/xmldom/assets/6949310/cda0ed5e-c632-4fad-a4fa-9d7a697cfad6

lib/grammar.js
homer0: Just in case, this isn't the plugin; The plugin ignores all comments that don't have a JSDoc tag (unless you enable [`jsdocExperimentalFormatCommentsWithoutTags`](https://github.com/homer0/packages/tree/main/packages/public/prettier-plugin-jsdoc#parse-comments-without-tags))

lib/conventions.js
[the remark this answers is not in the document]
homer0: Sorry, no, I didn't mean to say that you could do it "fix it" by changing the syntax; What I meant to say is that if you were to use `@throws {TypeError} If target is not an object.`, the type would stay next to the `@throws` instead of just moving everything to the next line:

```
 * @throws {TypeError}
 * If target is not an object.
```

Sorry for the confusion.

shunkica:
> > @karfau is there a reason yml files are used for config? Would you mind a switch to json? ( for prettier and eslint)
> 
> I'm not aware of any specific reason other than it being the default documented format?

Its one of the documented formats. I wouldn't say it is default.
I just find JSON easier to work with than some indentation based markup.
In any case It's not a big issue for me.

karfau (approved):
@shunkica I'm fine with the current state.

I configured the `jsdocPrintWidth` to 95, which is of course an arbitrary number, but I tried different values and this one looked best for the existing content as far as I could judge it.
(I did invest some time into formatting before this new mechanism, so even if it was not totally consistent it was also not without a any consideration.)

Adn I think I would prefer changing the config format in the next PR if you don't mind.

Let me know if there is anything you want ot add to this PR.

lib/conventions.js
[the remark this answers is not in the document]
homer0: @shunkica yeah, it doesn't seem like it would be too complicated: maybe two new options, to specify that the mode is enabled, and, the indentation value for the next lines.

Unfortunately, my current issue is time :P; I only have a few hours at night to work on personal stuff. I'll take note and add it to my "backlog", but I can't promise you that I'll make it a priority, sorry.

shunkica:
> @shunkica I'm fine with the current state.
> 
> I configured the `jsdocPrintWidth` to 95, which is of course an arbitrary number, but I tried different values and this one looked best for the existing content as far as I could judge it. (I did invest some time into formatting before this new mechanism, so even if it was not totally consistent it was also not without a any consideration.)
> 
> Adn I think I would prefer changing the config format in the next PR if you don't mind.
> 
> Let me know if there is anything you want ot add to this PR.

I stopped working yesterday because the IDE was was hammering my cpu to 250% yesterday and became unusable. I suspected the culprit was eslint/prettier but couldn't be bothered to investigate further at the time.

homer0:
> > @shunkica I'm fine with the current state.
> 
> > 
> 
> > I configured the `jsdocPrintWidth` to 95, which is of course an arbitrary number, but I tried different values and this one looked best for the existing content as far as I could judge it. (I did invest some time into formatting before this new mechanism, so even if it was not totally consistent it was also not without a any consideration.)
> 
> > 
> 
> > Adn I think I would prefer changing the config format in the next PR if you don't mind.
> 
> > 
> 
> > Let me know if there is anything you want ot add to this PR.
> 
> 
> 
> I stopped working yesterday because the IDE was was hammering my cpu to 250% yesterday and became unusable. I suspected the culprit was eslint/prettier but couldn't be bothered to investigate further at the time.

If you are using vscode: whenever you change Prettier/ESLint configs, or install/update dependencies, do a "reload window"; Linting tooling usually becomes buggy because it already had stuff in memory.

And just in case, no idea if you are doing it but, don't try to run eslint --fix with Prettier. ESLint is REALLY slow compared to Prettier, added to how fast you can be coding… it creates some _fun_ race conditions :P.

shunkica:
> If you are using vscode: whenever you change Prettier/ESLint configs, or install/update dependencies, do a "reload window"; Linting tooling usually becomes buggy because it already had stuff in memory.
> 
> And just in case, no idea if you are doing it but, don't try to run eslint --fix with Prettier. ESLint is REALLY slow compared to Prettier, added to how fast you can be coding… it creates some _fun_ race conditions :P.

I am actually using WebStorm, but I didn't have time to figure out what was going on.

karfau:
@shunkica let me know if you are still observing this slowness after a fresh computer/IDE start.

If yes, I will wait for you to figure out what is going on.

From the npm scripts we don't do eslint --fix, but I guess it is possible to invoke it from the IDE.

I think it is time to apply prettier to staged files in a pre-commit hook asap, to avoid running into formatting issues in the github actions. But this should be a separate PR.

Thx a lit @homer0 for all the support and for building and maintaining the plugin.

shunkica:
I can not reproduce the problems I had yesterday.

PR #514, opened 2023-07-16
fix(dom): prevent iteration over deleted items

### LINKED ISSUE #499: NodeList.item() returns removed nodes
**Example:**
```js
const doc = new DOMParser().parseFromString("<root><child/></root>", "text/xml")
const nodes = doc.getElementsByTagName("child");

// correct: { length: 1, item_0: "child" }
const first = {
    length: nodes.length,
    item_0: nodes.item(0)?.tagName ?? null,
};

doc.firstChild.removeChild(doc.firstChild.firstChild);

// should be: { length: 0, item_0: null}
// incorrect: { length: 0, item_0: "child" }
const second = {
    length: nodes.length,
    item_0: nodes.item(0)?.tagName ?? null,
};

console.log(first);
console.log(second);
```

**Cause:**
LiveNodeList is updated at [dom.js:312](https://github.com/xmldom/xmldom/blob/fbe1887b8d756c6b5bc419596240fec5c61c817a/lib/dom.js#L312) by copying all owned properties from an array. If the new array is shorter, indexes >= length are not removed.

**Possible Solution:**
Before or after `copy(ls, list)`, explicitly delete the extra properties:
```js
for (var i = ls.length; i in list; i += 1) {
    delete list[i];
}
```

**Runtime & Version:**
xmldom: 0.8.8
nodejs: 20.3.0

**Additional context:**
This causes problems with patterns like this:
```js
for (let node; (node = nodes.item(0)) !== null; node.parentNode.removeChild(node)) {
    // ...
}
```

karfau:
- `NodeList.item` now checks the `length` attribute before accessing the item at a provided index and returns `null` it the index is out of range
- when updating `LiveNodeList` (as part of accessing the `item` method) and the new `length` is shorter then the current one, any references on indexes following the index starting from `length` are `delete`d, and `null` is returned instead of `undefined`
- added some exports to the `lib/dom.js` module to enable testing

fixes #499

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -273,7 +273,7 @@ NodeList.prototype = {
 	 * 	The node at the indexth position in the NodeList, or null if that is not a valid index.
 	 */
 	item: function (index) {
-		return this[index] || null;
+		return index >= 0 && index < this.length ? this[index] : null;
 	},
 	toString: function (nodeFilter) {
 		for (var buf = [], i = 0; i < this.length; i++) {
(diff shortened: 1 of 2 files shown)


karfau:
@qtow from what I understood in your issue report, this should fix it in all the wyas I could think of.

Do you have a way to verify it before landing the PR? Or would you suggest to add any test? 
(I just made sure the changed lines are covered by tests.)

karfau:
The fix landed in 0.7 (lts), 0.8 (latest) and 0.9 (next)

PR #518, opened 2023-07-20
style: Translate/drop non English comments

shunkica:
I always wondered what this said.

### DIFF
--- a/lib/sax.js
+++ b/lib/sax.js
@@ -202,7 +202,7 @@ function parse(source, defaultNSMapCopy, entityMap, domBuilder, errorHandler) {
 		if (end > start) {
 			start = end;
 		} else {
-			//TODO: Possible sax fallback here, risk of positional error
+			//Possible sax fallback here, risk of positional error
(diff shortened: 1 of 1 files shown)


karfau (commented):
I already translated it multiple times, but never wanted to mix that change into the PR I was working on...
Did you try to understand if the code surrounding the comment makes sense? Or rather what the comment actually means?

I will check if that branch is covered by any test and if not try to find a sample that triggers it, to understand if it is actually something that we need TODO something about, or if it van just stand as an explanation of the branch.

karfau:
Even though I don't really understand the code in that place, it is covered by tests.
I would prefer to drop the "TODO" part of the comment, unless we know what we need to do here.

While we are at it, can you also translate the other comments that are not in English?

shunkica:
There is only one other non-English comment I could find and it says "忘记闭合" which translates to "forgot to close". Not very helpful :D

karfau:
in that case lets drop that comment

karfau (approved):

PR #519, opened 2023-07-22
fix!: report more non well-formed cases

### LINKED ISSUE #45: No error is being reported when the closing tag doesn't match the current start tag
I am wondering how I can detect errorneous input when using DOMParse.parseFromString(). To me it seems, that a errorneous element results in just ignoring it. 

Here is a simple demonstration:

**_errorneous.xml (missing '<' at the beginning of line 13):_**

```xml
<?xml version="1.0"?>
<catalog>
	<book id="bk101">
		<author>Gambardella, Matthew</author>
		<title>XML Developer's Guide</title>
		<genre>Computer</genre>
		<price>44.95</price>
		<publish_date>2000-10-01</publish_date>
		<description>An in-depth look at creating applications with XML.</description>
	</book>
	<book id="bk102">
		<author>Ralls, Kim</author>
		title>Midnight Rain</title>
		<genre>Fantasy</genre>
		<price>5.95</price>
		<publish_date>2000-12-16</publish_date>
		<description>A former architect battles corporate zombies, an evil sorceress, and her own childhood to become queen of the world.</description>
	</book>
</catalog>
```

**_parse.js:_**
```js
const fs = require("fs");
const DOMParser = require("xmldom").DOMParser;

let fileContent = fs.readFileSync("errorneous.xml", "utf-8");
let xmlDocument = new DOMParser().parseFromString(fileContent, "text/xml");

console.log(xmlDocument.getElementsByTagName("book").length);   // 2
console.log(xmlDocument.getElementsByTagName("title").length);  // 1
```

### LINKED ISSUE #125: Inconsistent attribute quote handling
I discovered this in a custom `assert.skip` call in `test/html/normalize.test.js`.

`new DOMParser().parseFromString('<r><Label onClick="doClick..>Hello, World</Label></r>','text/html')` (missing end quote in `onClick` attribute) leads to this with `&lt;`:

```
<r xmlns="http://www.w3.org/1999/xhtml">&lt;Label onClick="doClick..>Hello, World</r>
```

while `new DOMParser().parseFromString('<Label onClick=doClick..">Hello, World</Label>','text/html')` leads to this (looks correct):

```
<Label onClick="doClick.." xmlns="http://www.w3.org/1999/xhtml">Hello, World</Label>
```

### LINKED ISSUE #467: Inconsistent "span" behavior
When we have two spans next to each other and one of them is not enclosed properly, the later span text is added as a sibling of the span instead of as a child. This happens because of the space in `</span >`

If we enclose it properly it all works well. The thing is that formatters add tabs to html tags when the string is large.

A fix could be to use `html.replace(/[\t\s]+\>/g, ">")` to scape whitespace on enclosing tags, but it would be nice for this to be fixed inside xmldom.
**To Reproduce**
https://stackblitz.com/edit/js-xmldom-template-cvnj8l?file=index.js

**Expected behavior**
We want the span text to be a child of span.

**Runtime & Version:**
xmldom version: 0.8.6 
runtime version: node 16.14.2
other related software and version: Windows 10

karfau:
`fatalError` will throw a `ParseError` and stop processing the document:
- consistently detect current tag name
- report all possible cases of mismatching or unclosed tags in XML
  - case mismatches in HTML documents are not reported
- report missing attribute value in XML
- report unescaped `<` in AttrValue as fatalError (only in XML documents)
- report occurrences of Unicode replacement character `\ufffd`

`error` doesn't stop processing unless `onErrorStopParsing` or a similar method is used:
- report non white-space characters outside of elements (at the same level as the root element)

BREAKING-CHANGE: Reports more not well-formed documents as fatalError and drop broken support for optional and unclosed tags in HTML.

Closes #45 
Closes #125
Closes #467


test/html/__snapshots__/normalize.test.js.snap
```diff
@@ -378,6 +308,6 @@ Object {
 
 exports[`html normalizer unclosed html <html><meta><link><img><br><hr><input></html> 1`] = `
 Object {
-  "actual": "<html xmlns=\\"http://www.w3.org/1999/xhtml\\"><meta/><link/><img/><br/><hr/><input/></html>",
+  "actual": "<html xmlns=\\"http://www.w3.org/1999/xhtml\\"><meta><link><img><br><hr><input/></hr></br></img></link></meta></html>",
```
karfau, line 311: This is the most drastic/massive change that I might need to find a fix for before releasing 0.9.0, but to prevent this PR from growing indefinitely, it will have to be part of a later one.
Correct XML parsing has to be more important than the more lax/flexible HTML rules.

karfau: Solved in [`275693d` (#519)](https://github.com/xmldom/xmldom/pull/519/commits/275693d6727113c100b42fd4e7f2d345ff854a15)

test/error/reported.js
```diff
@@ -314,7 +441,13 @@
 			throw new Error(`line not mapped: ${lineKey} reportedAs $${key}`);
 		}
 	});
-	fs.writeFileSync(path.join(__dirname, 'reported.json'), JSON.stringify(LINE_TO_ERROR_INDEX, null, 2), 'utf8');
+
+	const REPORTED_JSON = path.join(__dirname, 'reported.json');
+	const data = JSON.stringify(LINE_TO_ERROR_INDEX, null, 2);
+	const currentData = fs.existsSync(REPORTED_JSON) ? fs.readFileSync(REPORTED_JSON, 'utf8') : '';
+	if (data !== currentData) {
+		fs.writeFileSync(REPORTED_JSON, data, 'utf8');
```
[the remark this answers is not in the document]
karfau: The additional check just avoids writing the file on every test run if the content doesn't change. It's not relevant for the published code

PR #550, opened 2023-10-04
fix: report fatalError when doctype is inside elements

karfau:
which is a regression compared to 0.8.x, most likely introduced as part of #498 

- add check to `parseDoctypeCommentOrCData`
- drop redundant and broken `Element.appendChild` implementation
- `hasInsertableNodeType` now checks for `CharacterData` nodes instead of only text nodes
- align ParseError and DOMException in how they are extending Error
- wrap `DOMException`s in `ParseError` in sax parser
- move custom errors to own module
  - and allow current and modern constructor arguments for DOMException


karfau (commented):
Self review completed

PR #552, opened 2023-10-06
xmldom optimacity

Maitrii1:
The issues were:

Missing MIME_TYPE constant
Using .toString() instead of an XML serializer
Appending the fragment before adding text


karfau:
I don't jnderstand what the purpose/intention of this pull request is

PR #634, opened 2024-03-16
feat: add iterator function to NodeList and NamedNodeMap

### LINKED ISSUE #633: [feature-request] Make `Element::childNodes` and `Element::attributes` iterable
**Is your feature request related to a problem? Please describe.**
It's currently not possible to iterate over the childNodes or attributes of an element like this:
```javascript
for (const child of element.childNodes) {
    console.log(child.nodeName);
}

for (const attr of element.attributes) {
    console.log(attr.value);
}
```
instead you have to write it like this:
```javascript
for (let i = 0; i < element.childNodes.length; i++) {
    const child = element.childNodes[i];
    console.log(child.nodeName);
}

for (let i = 0; i < element.attributes.length; i++) {
    const attr = element.attributes[i];
    console.log(attr.value);
}
```

**Describe the solution you'd like**
Add a generator function like this:
```javascript
[Symbol.iterator]: function * () {
    for (var i = 0; i < this.length; i++) {
        yield this[i];
    }
},
```
to `NodeList` and `NamedNodeMap`.

**Describe alternatives you've considered**
I am currently polyfilling this in my own project that uses xmldom so I can use `for...of` loops.

**Additional context**
I already implemented this and opened a PR [here](https://github.com/xmldom/xmldom/pull/634).

Ponynjaa:
Closes #633

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -120,15 +120,6 @@
 				"url": "https://opencollective.com/babel"
 			}
 		},
-		"node_modules/@babel/core/node_modules/semver": {
-			"version": "6.3.1",
-			"resolved": "https://registry.npmjs.org/semver/-/semver-6.3.1.tgz",
-			"integrity": "sha512-BR7VvDCVHO+q2xBEWskxS6DJE1qRnb7DxzUrogb71CWoSficBxYsiAGd+Kl0mmq/MprG9yArRkyrQxTO6XjMzA==",
-			"dev": true,
-			"bin": {
-				"semver": "bin/semver.js"
-			}
-		},
 		"node_modules/@babel/generator": {
 			"version": "7.23.0",
 			"resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.23.0.tgz",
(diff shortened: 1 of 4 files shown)


package-lock.json
karfau, line 1: Please revert the changes to the lock file.
Afterwards you can use `npm ci` to install the dependencies exactly as referenced in the lock file, without causing any unintentional updates.

karfau:
Can you please also add a test for these?
It doesn't match everywhere, but I'm trying to reflect the prototype function names in the lib/dom.js module by the files in the tests/dom folder.

karfau:
There currently is a restriction to only use ES5 inside the lib folder.
Is there a way to implement this with ES5?

Ponynjaa:
I added tests and rewrote the iterator to be ES5 compatible. As you can see it's way more code for the same thing but I guess it works...

karfau (approved):
I think when prefixing the unused `item` variables, that eslint complains about in the tests with `_` eslint should stop complaining.
Great job!

Ponynjaa:
@karfau done 😉

PR #637, opened 2024-03-18
feat: expose all DOM level 2 element prototypes

### LINKED ISSUE #40: Expose DOM Level 2 Elements

Ponynjaa:
Closes #40.

Exposes all fundamental and extended interfaces according to the [specs](https://www.w3.org/TR/2000/REC-DOM-Level-2-Core-20001113/core.html) while keeping the constructors for `Node`, `Document`, `Element`, `Attr`, `CharacterData`, `Text`, `Comment`,  `CDATASection`, `DocumentType`, `Notation`, `Entity`, `EntityReference`, `DocumentFragment`, `ProcessingInstruction` private by checking a symbol in the constructor.


Ponynjaa:
Also I forgot to apply the changes to some existing tests… I‘m gonna do that tomorrow 😅

karfau:
Great stuff, will review it in detail later, when the tests are passing.

Since we now have an approach for private constructors, there is no need for the DOM namespace.

The public constructors that are left now are not an issue from my perspective.

Regarding tests: I think I would prefer to have a check for each "class" in the related file. Can even be the first test in each file.

Is the symbol is even private for "us", meaning we don't even have access to it in the tests and have to use the DOMImplementation to create them?
I guess that makes the most sense, right?

A last remark: whatever is in the PR description usually goes into the commit message when squash merging. So we should avoid quwstions and instead describe the reasoning behind the changes. Questions can best be put in the comments. Would be nice if you can change that. Thx

PS: you don't need to mention me, I'm watching the repo for every change.

lib/errors.js
```diff
@@ -194,3 +194,30 @@ extendError(ParseError);
...
+	TYPE_MISMATCH_ERR: 17,
+	SECURITY_ERR: 18,
+	NETWORK_ERR: 19,
+	ABORT_ERR: 20,
+	URL_MISMATCH_ERR: 21,
+	QUOTA_EXCEEDED_ERR: 22,
+	TIMEOUT_ERR: 23,
+	INVALID_NODE_TYPE_ERR: 24,
+	DATA_CLONE_ERR: 25,
+};
```
karfau, line 223: Are they expected to be exported twice, once on DOMException and once on ExceptionCode?
And if so, can we at least refer to the values from DOMException here?

lib/conventions.js
```diff
@@ -141,6 +141,13 @@ var HTML_BOOLEAN_ATTRIBUTES = freeze({
 	selected: true,
 });
 
+/**
+ * Internal symbol used for construction of all classes whose constructors should be private.
+ * Currently used for checks in `Document`, `DocumentFragment`, `Node`, `Attr`, `Element`, `Text`, `CharacterData`
+ * so the constructor can't be used from outside the module.
+ */
+var INTERNAL_SYMBOL = Symbol('@xmldom/xmldom');
```
karfau, line 149: I think I would have called it something like `PRIVATE_DOM_CONSTRUCTOR` (maybe something shorter).

If we want to stick to being able to use it in tests, I would prefer to prefix the name with a `_` to indicate it is not part of the public API.

If we move it inside lib/dom.js and do not export it, the name doesn't matter as much, so it could just be `PDC`, which also saves some bytes at runtime.

Is it relevant what is passed to the Symbol constructor?
(I think conceptually I would think the symbol could be "owned" by the DOMImplementation, since it is the one calling those constructors, so maybe that's a better name to pass? But if it's totally private, an empty string could also be fine, right?)

lib/dom.js
```diff
@@ -2400,42 +2427,74 @@ Text.prototype = {
...
+	}
+}
 EntityReference.prototype.nodeType = ENTITY_REFERENCE_NODE;
 _extends(EntityReference, Node);
 
-function DocumentFragment() {}
+function DocumentFragment(symbol) {
+	if (symbol !== INTERNAL_SYMBOL) {
+		throw new Error('Illegal constructor');
+	}
```
karfau, line 2487: I think it's worth to extract this into a function on the module level which we pass the symbol to.

test/dom/attr.test.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+const { describe, test } = require('@jest/globals');
+const { Attr } = require('../../lib/dom');
+
+describe('Attr.prototype', () => {
+	describe('constructor', () => {
+		test('should throw Illegal constructor TypeError when trying to access constructor directly', () => {
+			expect(new Attr('attr')).toThrow(TypeError);
```
karfau, line 9: In order to use `.toThrow` you have to wrap the throwing code into a function, so it can be executed outside in a try catch as part of the assertion:
```suggestion
			expect(() => new Attr('attr')).toThrow(TypeError);
```

lib/errors.js
```diff
@@ -148,31 +148,40 @@ Object.defineProperties(DOMException.prototype, {
...
-DOMException.INDEX_SIZE_ERR = 1;
-DOMException.DOMSTRING_SIZE_ERR = 2;
-DOMException.HIERARCHY_REQUEST_ERR = 3;
-DOMException.WRONG_DOCUMENT_ERR = 4;
-DOMException.INVALID_CHARACTER_ERR = 5;
-DOMException.NO_DATA_ALLOWED_ERR = 6;
-DOMException.NO_MODIFICATION_ALLOWED_ERR = 7;
-DOMException.NOT_FOUND_ERR = 8;
-DOMException.NOT_SUPPORTED_ERR = 9;
-DOMException.INUSE_ATTRIBUTE_ERR = 10;
```
karfau, line 160: This value is used in some places, I even spotted one in the diff of one of the tests. How will this work?

Update: I just spotted the thing you do using `Object.entries` below, clever move. I'm curious to see if this works as expected...

Multiple contributors, including me, have been tempted to do something about this in the past.
Not sure if it is good to do it in this PR...

Ponynjaa: Is there a reason this should not work? 😄
From my tests it seems to work just as expected, but you can verify that, too - just to be sure.

karfau (approved):
Thx

PR #644, opened 2024-03-31
docs: click to copy install command in readme

jhauga:
Put install command in distinct block with triple backtick to easily copy/paste in order to save time (dozen seconds or so).


karfau (approved):
Thx

PR #652, opened 2024-04-18
feat: add DOM level 4 child element properties

dmitri-gb:
Hey!

Since https://github.com/xmldom/xmldom/pull/398 seems to be stale and apparently no longer being worked on by the original author, I figured I'd make another attempt.

This PR adds the following properties to `Element` nodes:
- `childElementCount`
- `firstElementChild`
- `lastElementChild`
- `nextElementSibling`
- `previousElementSibling`

These properties correspond to the attributes defined in the [`ParentNode`](https://dom.spec.whatwg.org/#interface-parentnode) and [`NonDocumentTypeChildNode`](https://dom.spec.whatwg.org/#interface-nondocumenttypechildnode) mixins in the DOM Living Standard.

The first three of these properties are also added to `Document` and `DocumentFragment`.

### DIFF
--- a/test/dom/dom-implementation.test.js
+++ b/test/dom/dom-implementation.test.js
@@ -62,6 +62,9 @@ describe('DOMImplementation', () => {
 			expect(root.prefix).toBe(null);
 			expect(root.localName).toBe(NAME);
 			expect(doc.documentElement).toBe(root);
+			expect(doc.childElementCount).toBe(1);
+			expect(doc.firstElementChild).toBe(root);
+			expect(doc.lastElementChild).toBe(root);
 			expect(doc.contentType).toBe(MIME_TYPE.XML_APPLICATION);
 			expect(doc.type).toBe('xml');
 		});
@@ -80,6 +83,9 @@ describe('DOMImplementation', () => {
 			expect(root.tagName).toBe(NAME);
 
 			expect(doc.documentElement).toBe(root);
+			expect(doc.childElementCount).toBe(1);
+			expect(doc.firstElementChild).toBe(root);
+			expect(doc.lastElementChild).toBe(root);
 			expect(doc.contentType).toBe(MIME_TYPE.XML_APPLICATION);
 			expect(doc.type).toBe('xml');
 		});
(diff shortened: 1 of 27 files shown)


test/dom/element.test.js
```diff
@@ -75,8 +75,8 @@ describe('documentElement', () => {
 		expect(doc.documentElement.toString()).toBe('<test>bye</test>');
 	});
 
-	test('appendElement and removeElement', () => {
-		const dom = new DOMParser().parseFromString(`<root><A/><B/><C/></root>`, MIME_TYPE.XML_TEXT);
+	test('appendChild and removeChild', () => {
+		const dom = new DOMParser().parseFromString(`<root><A/>x<B/>y<C/></root>`, MIME_TYPE.XML_TEXT);
```
karfau, line 79: I might need to check out the code to understand this change. Could we keep the existing test and add a new one for the new cases?
Or does it cover all 4 funtions, in that case Iwould prefer the message to cobtain all of them.

test/dom/dom-implementation.test.js
```diff
@@ -99,25 +105,9 @@ describe('DOMImplementation', () => {
 			expect(root.tagName).toBe(qualifiedName);
 
 			expect(doc.documentElement).toBe(root);
-			expect(doc.contentType).toBe(MIME_TYPE.XML_APPLICATION);
-			expect(doc.type).toBe('xml');
-		});
-
-		test('should create a Document with root element in a named namespace', () => {
```
karfau, line 106: Why was this dropped?

karfau (commented):
Thx for picking it up and investing into the tests. Did you read all the comments in the previous PR?
I think there was an idea to implement these functions in a "getter/setter" fashion that operates on the `childNodes` to avoid having to maintain redundant state... But maybe you already considered that, not sure, will need more time to understand the changes.

test/dom/dom-implementation.test.js
dmitri-gb: It was an exact duplicate of the test above.

test/dom/element.test.js
dmitri-gb: It only covers the two functions -- `appendChild` and `removeChild`. The test simply had the names wrong.

dmitri-gb:
> I think there was an idea to implement these functions in a "getter/setter" fashion that operates on the `childNodes` to avoid having to maintain redundant state...

I didn't spot any such idea in the previous PR. I'm not sure if that's really the way to go though. Accessing something as simple as `childElementCount` would then need to iterate over all child nodes and count how many of them are elements. These properties would then be very slow to read for very large DOM trees. I think maintaining extra state is probably the better option, no?

karfau:
> > I think there was an idea to implement these functions in a "getter/setter" fashion that operates on the `childNodes` to avoid having to maintain redundant state...
> 
> I didn't spot any such idea in the previous PR. I'm not sure if that's really the way to go though. Accessing something as simple as `childElementCount` would then need to iterate over all child nodes and count how many of them are elements. These properties would then be very slow to read for very large DOM trees. I think maintaining extra state is probably the better option, no?

It's always a trade between either keeping more state (in memory) and correctly updating all references when modifying the DOM (which also takes time and those things are happeing when ever a node is added during parsing), even when there might never be any access to that state, vs accessing the internal state when needed and have a (slightly) longer runtime for accessing those values.

with the amount of state maintenance in this lib I'm leaning towards doing as many things as possible using a getter/setter approach, but it has not been applied widely yet, and there might be issues with how "prototype inheritence" is implemented in this lib, which could get into the way of doing it like this, if it is implemented on prototypes that are copied.

Just thinking out loud here.

I would love to be more active on this repo, but this is only very likely to happen at the end of the year or start of the next year.
I hope this is not to big of an issue for you. Let me know when you think it's "done" so I can take a closer look whenever I have time.

dmitri-gb:
Hey, @karfau! Sorry for leaving this hanging for so long... I noticed that someone else recently implemented `children`, so I updated my implementation of the five `...Element...` properties to be based on `children`. There's no more separate state / bookkeeping for these properties.

I also needed to remove the `_extends` function (which already had a comment saying that the function didn't make sense), as the copying logic in it didn't work with getter properties. Inheritance is now done by directly setting the standardized [`__proto__`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#prototype_setter) and `constructor` properties inside the various `T.prototype` objects.

Additionally, I needed to update Jest to v30, as older versions are [unable to serialize and send cyclic structures to Jest workers](https://github.com/jestjs/jest/pull/15191) (the cyclic structures in this case are the DOM trees). This issue appeared after the inheritance-related change. I suspect that this may cause CI failures, as Jest v30 dropped support for Node.js v14 and v16.

PR #713, opened 2024-08-30
fix(types): DOMParser.parseFromString requires mimeType as second argument

krystofwoldrich:
Based on the spec and the current implementation the second argument of DOMParser.parseFromString is not optional and doesn't have a default value.

https://github.com/xmldom/xmldom/blob/9f9a74850a7c929dfd82462e9720a12097bbf426/lib/dom-parser.js#L215

https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString
https://html.spec.whatwg.org/#dom-domparser-parsefromstring-dev

This fixes types for changes made in https://github.com/xmldom/xmldom/releases/tag/0.9.0

Previous release https://github.com/xmldom/xmldom/releases/tag/0.8.10 allowed `mimeType` as optional.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -284,7 +284,7 @@ declare module '@xmldom/xmldom' {
 		 */
 		parseFromString(
 			source: string,
-			mimeType: MIME_TYPE = MIME_TYPE.XML_TEXT
+			mimeType: MIME_TYPE
 		): Document;
 	}
 


karfau (approved):
OMG, thx

karfau:
I will file a release within the next few days.

karfau:
@krystofwoldrich FYI: it will still take some time, since I want to avoid a ton of releases with just type fixes, so I'm working on #717 right now, which will fix all currently known type issues.

As part of that I also widened the accepted type to a literal type | string, just in case somebody only has that as a type for the mimeType to pass.

papandreou:
Would be really nice to get a new release out soon, as the latest version throws an error when you don't supply the `mimeType` arg, which makes it incompatible with things like `saml2-js`? That seems more important to fix than typings?

karfau:
@papandreou this is intentional breaking change in version 0.9.0 which aligns the behavior of xmldom with the specs. This is not going to be "fixed", since it works as expected.
If the previous behavior is required you can not upgrade and have to stick to 0.8.*! 
I do not recommend sticking to this behavior, since it means the content that you are parsing is used to determine the mime type. Which is only secure if you know what kind of content you expect, in which case you can also put the mime type as an argument. 

The issue described here, is that this change is not reflected in the types of 0.9.0, which will be fixed along with all other type issues as part of the PR linked in my previous comment.

for further discussion on this topic I would prefer to have it in the release discussion #435

papandreou:
Ah, apologies, I thought it was the other way around 🙈

karfau:
In 0.*  (aka unstable) versions, breaking changes are happening in minor versions.
There are tools like Renovate, that are aware  is aware of this. 
Sorry that you were not aware of it, and were affected by this.

PR #721, opened 2024-09-05
fix: add missing type definitions

### LINKED ISSUE #720: TS: Missing types for `Document.documentElement` and `Element.tagName`
**Describe the bug**
Trying to access `document.documentElement` or `Element.tagName` with TypeScript gives a type error, even though these properties are supported.

**To Reproduce**
```ts
import { DOMImplementation } from '@xmldom/xmldom';

const doc = new DOMImplementation().createDocument(null, 'qualifiedName');
const documentElement = doc.documentElement;
// Property 'documentElement' does not exist on type 'Document'.

const element = doc.createElement('element');
const tagName = element.tagName;
// Property 'tagName' does not exist on type 'Element'.
```

**Expected behavior**
`documentElement` should be available with the type `Element` or `Element | null` and `tagName` should be available with the type `string`

**Runtime & Version:**
xmldom version: 0.9.1

Mathias-S:
Adds types for properties that are exported,
but are missing in index.d.ts:

- `Document.domElement`
- `Element.tagName`

Fixes #720

### DIFF
--- a/examples/typescript-node-es6/src/index.ts
+++ b/examples/typescript-node-es6/src/index.ts
@@ -23,6 +23,7 @@ import {
 	ParseError,
 	Text,
 	XMLSerializer,
+	Element,
 } from '@xmldom/xmldom';
 
 const failedAssertions: Error[] = [];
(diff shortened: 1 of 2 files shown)


karfau (approved):

karfau:
Thx a lot!

PR #728, opened 2024-09-10
Add missing node props

### LINKED ISSUE #724: Argument of type 'Document' is not assignable to parameter of type 'Node' after update to xmldom 0.9.2
**Describe the bug**
It looks like there is a regression issue related to type definitions since release 0.9.2. The actual error message is `Argument of type 'Document' is not assignable to parameter of type 'Node'.
  Type 'Document' is missing the following properties from type 'Node': baseURI, isConnected, parentElement,`

Before xmldom 0.9.2 patch there were no linter errors detected in the same piece of code.

**To Reproduce**
```
import {DOMParser, MIME_TYPE} from '@xmldom/xmldom';

function runQuery(query, xmlStr) {
  const dom = new DOMParser().parseFromString(xmlStr, MIME_TYPE.XML_TEXT);
  const nodes = xpathQuery(query, dom);
  return nodes;
}
```

**Expected behavior**
No linter errors

**Runtime & Version:**
xmldom version: 0.9.2
runtime version: any
other related software and version: 

**Additional context**

https://github.com/appium/appium/pull/20537/commits/67799c96da568c9b8eac9b3ec787939771a0c478

Ponynjaa:
Adds missing properties and methods of Node to be compatible with the default DOM-typings.

Closes #724

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -350,7 +350,7 @@ declare module '@xmldom/xmldom' {
 	 * cannot have children will throw an exception.
 	 *
 	 * **This behavior is slightly different from the in the specs**:
-	 * - undeclared properties: nodeType, baseURI, isConnected, parentElement, textContent
+	 * - undeclared properties: baseURI, isConnected, parentElement
 	 * - missing methods: contains, getRootNode, isEqualNode, isSameNode
 	 *
 	 * @see http://www.w3.org/TR/2000/REC-DOM-Level-2-Core-20001113/core.html#ID-1950641247
@@ -396,6 +396,12 @@ declare module '@xmldom/xmldom' {
 		 * [MDN Reference](https://developer.mozilla.org/docs/Web/API/Node/nodeName)
 		 */
 		readonly nodeName: string;
+		/**
+		 * Returns the type of node.
+		 *
+		 * [MDN Reference](https://developer.mozilla.org/docs/Web/API/Node/nodeType)
+		 */
+		readonly nodeType: number;
 		/** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Node/nodeValue) */
 		nodeValue: string | null;
 		/**
@@ -420,6 +426,19 @@ declare module '@xmldom/xmldom' {
 		 * [MDN Reference](https://developer.mozilla.org/docs/Web/API/Node/previousSibling)
 		 */
 		readonly previousSibling: Node | null;
+		/** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Node/textContent) */
+		textContent: string | null;
+
+		/**
+		 * Zero based line position inside the parsed source,
+		 * if the `locator` was not disabled.
+		 */
+		lineNumber?: number;
+		/**
+		 * One based column position inside the parsed source,
+		 * if the `locator` was not disabled.
+		 */
+		columnNumber?: number;
 
 		/** [MDN Reference](https://developer.mozilla.org/docs/Web/API/Node/appendChild) */
 		appendChild(node: Node): Node;
@@ -1061,6 +1080,13 @@ declare module '@xmldom/xmldom' {
 
 	interface ProcessingInstruction extends Node {
 		nodeType: typeof Node.PROCESSING_INSTRUCTION_NODE;
+		/**
+		 * Everything that goes after the target, excluding `?>`.
+		 * [MDN Reference](https://developer.mozilla.org/docs/Web/API/CharacterData/data)
+		 */
+		data: string;
+		/** [MDN Reference](https://developer.mozilla.org/docs/Web/API/ProcessingInstruction/target) */
+		readonly target: string;
 	}
 	var ProcessingInstruction: InstanceOf<ProcessingInstruction>;
 
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -979,7 +979,7 @@ DOMImplementation.prototype = {
  * cannot have children will throw an exception.
  *
  * **This behavior is slightly different from the in the specs**:
- * - undeclared properties: nodeType, baseURI, isConnected, parentElement, textContent
+ * - undeclared properties: baseURI, isConnected, parentElement
  * - missing methods: contains, getRootNode, isEqualNode, isSameNode
  *
  * @class
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -978,10 +978,6 @@ DOMImplementation.prototype = {
  * depending on circumstances. For example, attempting to add children to a node type that
  * cannot have children will throw an exception.
  *
- * **This behavior is slightly different from the in the specs**:
- * - undeclared properties: baseURI, isConnected, parentElement
- * - missing methods: contains, getRootNode, isEqualNode, isSameNode
- *
  * @class
  * @abstract
  * @param {Symbol} symbol
@@ -1025,6 +1021,12 @@ Node.prototype = {
 	 */
 	parentNode: null,
 	/**
+	 * The parent element of this node.
+	 *
+	 * @type {Element | null}
+	 */
+	parentElement: null,
+	/**
 	 * The child nodes of this node.
 	 *
 	 * @type {NodeList | null}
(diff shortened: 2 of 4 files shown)


Ponynjaa:
FYI: I'm not done yet. I only added the logic and some descriptions to the props/methods that I added however I didn't have time yet to update the README, add typings or add tests.
If you want to take over from here feel free @karfau.

lib/dom.js
```diff
@@ -1919,12 +2067,26 @@ function _insertBefore(parent, node, child, _inDocumentAssertion) {
 	}
 	do {
 		newFirst.parentNode = parent;
+		if (parent.nodeType === parent.ELEMENT_NODE) {
+			newFirst.parentElement = parent;
+		} else {
+			newFirst.parentElement = null;
+		}
```
karfau, line 2074: I think it would be better to implement this as a getter, to avoid having to maintain this value when the DOM structure is changed.

lib/dom.js
```diff
@@ -1919,12 +2067,26 @@ function _insertBefore(parent, node, child, _inDocumentAssertion) {
...
+
+	// find out if node is connected to a document
+	var parentNode = parent;
+	do {
+		if (parentNode.nodeType === parentNode.DOCUMENT_NODE) {
+			node.isConnected = true;
+			break;
+		}
+	} while ((parentNode = parentNode.parentNode));
+
```
karfau, line 2089: Same here regarding implementing it as a getter, since this can be quite some iterations in big documents even during parsing, and the property would also need to be maintained as part of other operations for every node in the tree, even when it is never requested.

And I think the implementation should just use the new `getRootNode` implementation and check it's `nodeType`

readme.md
```diff
@@ -317,23 +317,23 @@ The original author claims that xmldom implements [DOM Level 2] in a "fully comp
...
+- `Element`
 - `Entity`
 - `EntityReference`
 - `LiveNodeList`
-- `NamedNodeMap` *
-- `Node` *
-- `NodeList` *
+- `NamedNodeMap`
+- `Node`
+- `NodeList`
```
karfau, line 336: Good catch, I even found more things that are outdated, like the node version... I guess we should review the whole document.
But it doesn't have to happen in this pull request.

karfau (commented):
I just had a first glance at the diff, but already wanted to share the feedback

Ponynjaa:
graphic
Haha all this might have been for nothing 😢

karfau:
> graphic Haha all this might have been for nothing 😢

I don't think so. 
It's a step towards a more complete implementation, but I never expected that we will implement all properties/methods, unless we clutter the code with a lot of "NotImplemented" methods.

One possible approach would be to vendor the whole `lib.dom.d.ts` and mark all methods that are not implemented as deprecated ... but that would again mean that type checks would pass that shouldn't.

But I think we will need some kind of helper to make it possible/easy to tell typescript it should be treated as compatible. Maybe by using some generic types?
Something on the level of DocumentImplementation that is passed down to all the methods of the document it creates, so that the global types can be passed as type argument?
Will have to experiment with it.

index.d.ts
```diff
@@ -410,6 +429,12 @@ declare module '@xmldom/xmldom' {
 		 * [MDN Reference](https://developer.mozilla.org/docs/Web/API/Node/parentNode)
 		 */
 		readonly parentNode: Node | null;
+		/**
+		 * Returns the parent element.
```
karfau, line 433: I would like to make it more explicit that it is just checking the direct parent, e.g.

> Returns the parent `Node` if it is of type `Element`, otherwise `null`.

lib/dom.js
```diff
@@ -1060,6 +1068,167 @@ Node.prototype = {
...
+	 * @param {Node} [other]
+	 * The node to check.
+	 * @returns {boolean}
+	 * True if `other` is an inclusive descendant of this node; false otherwise.
+	 * @see https://dom.spec.whatwg.org/#dom-node-contains
+	 */
+	contains: function (other) {
+		if (!other) {
+			return false;
+		}
```
karfau, line 1105: I would suggest we add a test that passes `undefined` or `null` to the function to cover the condition.
It makes sense to have it, even though the types say you can only pass a `Node` which can never be falsy.

As long as our tests are not type checked, it is possible to cover the case :)

lib/dom.js
```diff
@@ -1060,6 +1068,167 @@ Node.prototype = {
...
+	 * options.compose doesn't change this behaviour.
+	 *
+	 * @param {GetRootNodeOptions} [options]
+	 * @returns {Node}
+	 * Root node.
+	 * @see https://dom.spec.whatwg.org/#dom-node-getrootnode
+	 */
+	getRootNode: function (options) {
+		if (options && options.composed) {
+			// should return shadow root here (not supported)
```
karfau, line 1132: Question: Should we rather throw `new Error('not implemented: composed === true');`, instead of returning the wrong value? Or even go the extra mile of implementing that case?

lib/dom.js
```diff
@@ -1060,6 +1068,167 @@ Node.prototype = {
...
+			return false;
+		}
+
+		if (this.childNodes && otherNode.childNodes) {
+			for (var i = 0; i < this.childNodes.length; i++) {
+				if (!this.childNodes[i].isEqualNode(otherNode.childNodes[i])) {
+					return false;
+				}
+			}
+		}
```
karfau, line 1171: According to https://dom.spec.whatwg.org/#concept-node-equals and since it's the slowest possible check (since it recurses through all childnodes), it should go last in the function.

And we need at least one more test case as indicated by codecov.

lib/dom.js
```diff
@@ -1060,6 +1068,167 @@ Node.prototype = {
...
+
+		if (this.nodeType !== otherNode.nodeType) {
+			return false;
+		}
+
+		if (
+			(this.childNodes && otherNode.childNodes && this.childNodes.length !== otherNode.childNodes.length) ||
+			(!this.childNodes && otherNode.childNodes) ||
+			(this.childNodes && !otherNode.childNodes)
```
karfau, line 1160: All `Node`s have a `childNodes` attribute, right? How can it not be set?

I think it is because we do not really have classes and inheritance so the property is not set from all the "constructors"...

We could fix it in this PR, by setting `childNodes` in all contructors, with the benefit of simpifying this check to
`this.childNodes.length !== otherNode.childNodes.length`,
or we keep it for another PR and will file an issue.

What do you think?

lib/dom.js
```diff
@@ -1060,6 +1068,167 @@ Node.prototype = {
...
+					this.prefix !== otherNode.prefix ||
+					this.localName !== otherNode.localName ||
+					(this.attributes && otherNode.attributes && this.attributes.length !== otherNode.attributes.length) ||
+					(!this.attributes && otherNode.attributes) ||
+					(this.attributes && !otherNode.attributes)
+				) {
+					return false;
+				}
+
+				if (this.attributes && otherNode.attributes) {
```
karfau, line 1191: But I think for `Elements`s we know that attributed can not be falsy, so here it should be possible to simplify the conditions.
I would even recommend to split the first condition into multiple short if conditions that each return `false`,
the last one being
`this.attributes.length !== otherNode.attributes.length`.
The one around the for loop should not be needed at all.

karfau (commented):
Completed my review, once the comments have been resolved, I will approve it.

karfau:
By the way, collaboration would be much easier for me if you push the branch to the xmldom repository, which you should be able to.

lib/dom.js
karfau: Just looked it up, since we do not have/support `ShadowRoot`s so far that would be a bit to much.
and also even when `composed === true`, it would only return the shadow root if the root is a shadow root, otherwise it would return the normal root. So throwing doesn't make sense.

I will tweak the comment accordingly.

lib/dom.js
karfau: FYI: I added `childNodes = new NodeList()` for every `Document.create*() => Node` method, but kept the condition for now, we will see if it's possible to produce test cases for all the missing cases.

karfau:
I resolved most of my comments, but there are some uncovered cases, which we need to add tests for.
They appear on the related lines when reviewing the files.
I will try to pick up one after the other during the week, unless you are faster :)

Ponynjaa:
I agree, we should just remove the checks for `.nodeList` on `isEqualNode` since it should always be there. It's too bad we can't just add this to the Node constructor, since it's easy to miss this on some "child classes" (or in our case the `createXXX` methods), I actually found one method that you probably missed, it was `createDocumentType`.
But with this out of the way it should be fine to just remove the checks so I don't have to write any more tests for these edge cases... there are already like 20 tests just for the `isEqualNode` function alone 😅 
But yeah this should be fine for the release now I think :)

karfau:
> Btw I have no idea why CodeCov wanted tests for [this added line](https://github.com/xmldom/xmldom/pull/728/commits/13d91ef706b21e4c24bb536a72025347defe099e#diff-79db8c65cf49002eb2d5d7e81a17f3b73d94db6bb50c0c12f0c9f427aff4b11eR2249)

Because there are no tests for `Document.createEntityReference` so far and we changed it :)

Even though codecov seems to be happy (I need to see how to configure this better), I would prefer that all changed code is covered before the PR is merged.

karfau (approved):
Thx, great contribution and collaboration

PR #766, opened 2024-10-23
fix: re-index childNodes on insertBefore

### LINKED ISSUE #763: insertBefore is appending the node in childNodes (0.9.4)
**Describe the bug**
When manipulating a document object with processing instructions, it should be possible to add a doctype using `insertBefore`. This used to work up until 0.9.3, and was broken in 0.9.4, most probably by #760 

**To Reproduce**
I was not able to reproduce this behavior on stackblitz, but see expected behavior below with a unit test that does reproduce it

**Expected behavior**
Add the following test to `document.test.js`, under "insertBefore":
```javascript
test('should insert doctype when processing instruction is present', () => {
	const doc = new DOMImplementation().createDocument(null, '');
	expect(doc.childNodes).toHaveLength(0);
	expect(doc.documentElement).toBeNull();

	const instruction = doc.createProcessingInstruction('target', 'data');
	doc.insertBefore(instruction);

	const root = doc.createElement('root');
	doc.appendChild(root);
	expect(doc.childNodes).toHaveLength(2);
	expect(doc.childNodes.item(0)).toBe(instruction);
	expect(doc.childNodes.item(1)).toBe(root);

	const doctype = doc.implementation.createDocumentType('qualifiedName', '', '');
	doc.insertBefore(doctype, root);
	expect(doc.childNodes).toHaveLength(3);
	expect(doc.childNodes.item(0)).toBe(instruction);
	expect(doc.childNodes.item(1)).toBe(doctype); // This assertion fails in 0.9.4, but passes in 0.9.3
	expect(doc.childNodes.item(2)).toBe(root);
});
```

This test passes when applied on the 0.9.3 tag and fails on the 0.9.4 tag

**Runtime & Version:**
xmldom version: 0.9.4
runtime version: Tested with Node.js 18.0.0, 20.15.1 and 22.3.0
other related software and version:  n/a

**Additional context**

karfau:
to resolve a regression introduced in #760

fixes #763

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -1623,30 +1623,29 @@ function _onRemoveAttribute(doc, el, newAttr, remove) {
 }
 
 /**
- * Updates `el.childNodes`, adjusting the indexed items and its `length`.
- * If `newChild` is provided, it will be appended to the childNodes list.
- * Otherwise, it's assumed that an item has been removed,
- * and `el.firstNode` and its `.nextSibling` are used to iterate over the current list of child
- * nodes, effectively reindexing them.
+ * Updates `parent.childNodes`, adjusting the indexed items and its `length`.
+ * If `newChild` is provided and has no nextSibling, it will be appended.
+ * Otherwise, it's assumed that an item has been removed or inserted,
+ * and `parent.firstNode` and its `.nextSibling` to re-indexing all child nodes of `parent`.
  *
  * @param {Document} doc
  * The parent document of `el`.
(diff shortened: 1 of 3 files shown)


mureinik:
FWIW, I can confirm that this PR solves not only the issue in #760, but also the original issue I faced in my work which #760 is a reduction thereof.

karfau:
That's good to know, I hope to finish this up and release it latest on Monday

PR #803, opened 2024-12-11
fix: locator is now true even when other options are being used for the DOMParser

### LINKED ISSUE #802: locator isn't true by default when using any other options
**Describe the bug**
When providing options to the `DOMParser` constructor and leaving out the `locator` option this option will not be set to true even though the typedef says it is true by default.

It is only being set to true [here](https://github.com/xmldom/xmldom/blob/a561f539b928ab0e641758098a74e03178d0b55a/lib/dom-parser.js#L104) when the options are omitted completely.

**To Reproduce**
```js
import * as xmldom from '@xmldom/xmldom';

const parser = new xmldom.DOMParser({
	onError: (level, message) => {
		console.log({level, message});
	}
});

try {
	const doc = parser.parseFromString(`<?xml version="1.0" encoding="utf-8"?>
		<content>
			<chapter>
		</content>
	`, 'text/xml');
} catch (error) {
	console.log(error.locator); // -> undefined
}
```

**Expected behavior**
In the example above `error.locator` should be `{ lineNumber: 3, columnNumber: 13 }`.

**Runtime & Version:**
xmldom version: 0.9.2
runtime version: Node 20

Ponynjaa:
Fixed locator not being true by default when using any other options.
Closes #802

### DIFF
--- a/lib/dom-parser.js
+++ b/lib/dom-parser.js
@@ -101,7 +101,10 @@ function normalizeLineEndings(input) {
  * @see https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#dom-parsing-and-serialization
  */
 function DOMParser(options) {
-	options = options || { locator: true };
+	options = options || {};
+	if (typeof options.locator === 'undefined') {
+		options.locator = true;
+	}
 
 	/**
 	 * The method to use instead of `conventions.assign`, which is used to copy values from
--- a/lib/dom-parser.js
+++ b/lib/dom-parser.js
@@ -102,7 +102,7 @@ function normalizeLineEndings(input) {
  */
 function DOMParser(options) {
 	options = options || {};
-	if (typeof options.locator === 'undefined') {
+	if (options.locator === undefined) {
 		options.locator = true;
 	}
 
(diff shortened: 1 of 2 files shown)


Ponynjaa:
Hi @karfau ,
I'm a little surprised about the test that's now failing... it seems like this was the intended behavior before my change... however why is it documented that the default value of this option is true, when it is not? This seems like a mistake for me...
what do you think?

lib/dom-parser.js
```diff
@@ -101,7 +101,10 @@ function normalizeLineEndings(input) {
  * @see https://html.spec.whatwg.org/multipage/dynamic-markup-insertion.html#dom-parsing-and-serialization
  */
 function DOMParser(options) {
-	options = options || { locator: true };
+	options = options || {};
+	if (typeof options.locator === 'undefined') {
```
karfau, line 105: How is this different from
```suggestion
	if (options.locator === undefined) {
```
?

karfau (commented):
Thank you for reporting and providing a fix.
I think the bug report and fix make sense.
We just need to change the expectation in the failing test.

karfau:
Regarding the breaking change, I would say this was more an oversight. And the way the docs are describing it makes more sense.
And since it is already the default when no options are passed and it is easy to explicitly turn it of if you already have options i would publish it as a patch change.

lib/dom-parser.js
Ponynjaa: In this case it doesn't make any difference, so yeah I'll change that

karfau (approved):

PR #805, opened 2024-12-11
perf: simplify `DOM.compareDocumentPosition`

zorkow:
PR contributes a more efficient coding for the `compareDocumentPosition` method contributed in PR #488 
. In detail:

* Removes the `parentChain` method and replaces it with a parent chain computation that immediately compares target nodes, thus avoiding unnecessary computations.
* Uses `push` and `reverse` instead of `unshift` to assemble chains as this is considerably faster.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -258,23 +258,6 @@ var DocumentPosition = conventions.freeze({
 
 //helper functions for compareDocumentPosition
 /**
- * Constructs a parent chain for a node.
- *
- * @param {Node} node
- * The start node from which the parent chain will be constructed.
- * @returns {Node[]}
- * The array of nodes representing the parent chain from the root to the specified node.
- */
-function parentChain(node) {
-	var chain = [];
-	while (node.parentNode || node.ownerElement) {
-		node = node.parentNode || node.ownerElement;
-		chain.unshift(node);
-	}
-	return chain;
-}
-
-/**
  * Finds the common ancestor in two parent chains.
  *
  * @param {Node[]} a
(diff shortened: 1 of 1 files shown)


karfau (approved):

karfau:
Thx again, it even looks faster :wink:

When you update the branch against the latest master branch, I can merge it.

zorkow:
> When you update the branch against the latest master branch, I can merge it.

Done. 

On a different note: Are you considering moving to ES modules at some point. I did a quick and dirty mockup the other day for the MathJax dev lab, where modules are imported directly from source rather than bundled, and conversion took about ten minutes.

karfau:
> On a different note: Are you considering moving to ES modules at some point. I did a quick and dirty mockup the other day for the MathJax dev lab, where modules are imported directly from source rather than bundled, and conversion took about ten minutes.

I'm really low on resources regarding this project currently, so I will very likely not be the person to initiate this.
I also guess it would be another breaking change and what to do about still supporting the current ways of using the lib, ... I did not even have the time to fully understand the available options and what it would take.
But if somebody else makes it happen in a way that is possible for me to sustain/maintain I will consider it.
The technical overhaul of this lib is overdue, but my capacity is limited and there are multiple other options out there that might be a better fit for those use cases...

zorkow:
Thanks for the merge.

> I'm really low on resources regarding this project currently, so I will very likely not be the person to initiate this.

I know what you mean. It's not as if I had any free cycles. It would be a shame to see that project die, though, as it really is a lean and effective DOM Implementation. There are many bloated ones around... 
Anyway, for maybe my es6 hack will help someone in the future:
https://github.com/Speech-Rule-Engine/xmldom/tree/feature/es6_modules/mjs

As I said quick and dirty, mostly done with a couple of `sed` scripts.

PR #839, opened 2025-02-18
fix: replace \u2029 as part of normalizeLineEndings

### LINKED ISSUE #838: Increased processing times when using \u2028 and \u2029
**Describe the bug**
If the content contains an `\u2029` or `\u2028`, the processing of the XML takes considerably longer. The reason for this is the `Regex::exec` function in connection with `.*`. 

The reason for this is the following regular expression: 
https://github.com/xmldom/xmldom/blob/master/lib/sax.js#L112

**To Reproduce**
The error can also be reproduced relatively easily without XML Dom using the following snippet: 

```js
var contents = [
	`<div>${"A".repeat(50000)}\u2029${"B".repeat(10)}</div>`,
	`<div>${"A".repeat(50000)}\n${"B".repeat(10)}</div>`
];

const patterns = [
	/.*(?:\r\n?|\n)|.*$/g,
	/.*(?:\r\n?|[\n\u2028\u2029])|.*$/g
];


for (const pattern of patterns) {
	console.log("================");
	console.log("  Pattern", pattern);


	for (let i = 0, m = 2; i < m; i++) {
		const startTime = Date.now();	

		let linePattern = pattern;
		let content = contents[i];
		linePattern.lastIndex = -1;

		let m;
		let count = 0;
		while ((m = linePattern.exec(content))) {
			if (m.index === linePattern.lastIndex) {
				linePattern.lastIndex++;
			}
			count++;
		}
			
		console.log("    Matcher", i, "found" ,  count," and took ", (Date.now() - startTime) / 1000, "sec");
	}
}
```

This will print the following output:

```bash
================
  Pattern /.*(?:\r\n?|\n)|.*$/g
    Matcher 0 found 2  and took  4.118 sec
    Matcher 1 found 3  and took  0 sec
================
  Pattern /.*(?:\r\n?|[\n\u2028\u2029])|.*$/g
    Matcher 0 found 3  and took  0 sec
    Matcher 1 found 3  and took  0 sec
```

To reproduce it in Stackblitz i created the following Stackblitz:
https://stackblitz.com/edit/js-xmldom-template-cmzfeewj?file=index.js,package.json

**Expected behavior**
It should not take longer just because a `\u2028` or `\u2029` exists in the document

**Runtime & Version:**
xmldom version: 0.9.7
runtime version: Browser or NodeJS
other related software and version: - 

**Additional context**
A PR for the customisation of the RegExp will follow.

kboshold:
Fixes #838, when the default `normalizeLineEndings` implementation is being used

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1527,12 +1527,15 @@ declare module '@xmldom/xmldom' {
 		readonly locator?: boolean;
 
 		/**
-		 * used to replace line endings before parsing, defaults to `normalizeLineEndings`,
-		 * which normalizes line endings according to <https://www.w3.org/TR/xml11/#sec-line-ends>.
+		 * used to replace line endings before parsing, defaults to exported `normalizeLineEndings`,
+		 * which normalizes line endings according to <https://www.w3.org/TR/xml11/#sec-line-ends>,
+		 * including some Unicode "newline" characters.
+		 *
+		 * @see {@link normalizeLineEndings}
 		 */
 		readonly normalizeLineEndings?: (source: string) => string;
 		/**
-		 * A function that is invoked for every error that occurs during parsing.
+		 * A function invoked for every error that occurs during parsing.
 		 *
 		 * If it is not provided, all errors are reported to `console.error`
 		 * and only `fatalError`s are thrown as a `ParseError`,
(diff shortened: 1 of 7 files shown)


Ponynjaa (approved):

karfau (approved):
Awesome work, thx a lot

PR #846, opened 2025-02-25
fix: fixed duration check in alternative whitespace test

Ponynjaa:
Fixed duration check in alternative whitespace test not working since jest doesn't measure execution time within sync functions.
Added manual check for execution time.


karfau:
@Ponynjaa Just read about your PR, but my approach to enforce the timeout is similar in #847 and includes a fix, so I guess we can close this one?

PR #847, opened 2025-02-25
perf: improve line detection regular expression for locator

karfau:
- enforce timeouts on tests and specifically for the one not using the default `normalizeLineEndings`, so it actually fails.

### DIFF
--- a/test/dom-parser.test.js
+++ b/test/dom-parser.test.js
@@ -1,6 +1,7 @@
 'use strict';
 
-const { describe, expect, test } = require('@jest/globals');
+const { performance } = require('node:perf_hooks');
+const { describe, expect, test, beforeEach, afterEach, afterAll } = require('@jest/globals');
 const { DOMParser, XMLSerializer } = require('../lib');
 const { assign, MIME_TYPE, NAMESPACE } = require('../lib/conventions');
 const { __DOMHandler, onErrorStopParsing, onWarningStopParsing } = require('../lib/dom-parser');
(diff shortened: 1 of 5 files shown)


karfau:
@kboshold this one is an actual fix for the performance issue from your initial example, even when not normalizing the input.
Also the timeout of the test is now enforced.

test/dom-parser.test.js
```diff
@@ -330,7 +331,20 @@ describe('DOMParser', () => {
 			const source = `<root>${'A'.repeat(50000)}\u2029${'A'.repeat(50000)}\u0085${'A'.repeat(50000)}\u2028${'A'.repeat(50000)}\u2029</root>`;
 			const doc = parser.parseFromString(source, MIME_TYPE.XML_TEXT);
 			expect(new XMLSerializer().serializeToString(doc)).toEqual(source.replace(/[\u0085\u2028\u2029]/g, '\n'));
-		}, 500);
+		});
+		test('should be able to open documents with alternative whitespace without creating a bottleneck and replacing them with \\n', () => {
```
Ponynjaa, line 335: > and replacing them with \\n

this description doesn't make sense for this test

test/dom-parser.test.js
```diff
@@ -330,7 +331,20 @@ describe('DOMParser', () => {
 			const source = `<root>${'A'.repeat(50000)}\u2029${'A'.repeat(50000)}\u0085${'A'.repeat(50000)}\u2028${'A'.repeat(50000)}\u2029</root>`;
```
Ponynjaa, line 331: > without creating a bottleneck

The test description here actually also doesn't make sense, maybe you can rephrase it when you are already at it?

Ponynjaa (changes requested):
Just two minor things I would change: the test-descriptions on the test you copied and the new one don't describe what's actually being tested.

Other than that it looks really good! I was a little confused at first when I saw your changes to the regex and loop, but after playing around with it for a little bit and testing the outcomes it seems to work exactly as before but without matching the content inbetween line-endings for no reason. Good job dude!

test/dom-parser.test.js
karfau: Yes, somehow the test I copied is also covering the one that @kboshold added, so I merged them into one and corrected the description.

test/dom-parser.test.js
```diff
@@ -330,7 +331,20 @@ describe('DOMParser', () => {
...
 			expect(new XMLSerializer().serializeToString(doc)).toEqual(source.replace(/[\u0085\u2028\u2029]/g, '\n'));
-		}, 500);
+		});
+		test('should be able to open documents with alternative whitespace without creating a bottleneck and replacing them with \\n', () => {
+			// issue: https://github.com/xmldom/xmldom/issues/838
+			const start = performance.now();
+			const onError = jest.fn();
+			const normalizeLineEndings = jest.fn((source) => source);
+			const { parser } = getTestParser({ onError, normalizeLineEndings });
+			const source = `<root>${'A'.repeat(15000)}\u2029${'A'.repeat(15000)}\u0085${'A'.repeat(15000)}\u2028${'A'.repeat(15000)}\u2029</root>`;
```
kboshold, line 333: I think it would make more sense to have a `\n` as the first character, since the test currently has `\u2029` twice.

Solution:
```js
const source = `<root>${'A'.repeat(15000)}\n${'A'.repeat(15000)}\u0085${'A'.repeat(15000)}\u2028${'A'.repeat(15000)}\u2029</root>`;
```

kboshold (changes requested):

karfau:
I applied all the suggested changes, so please approve it.
I would say the PR is already an improvement that we should get landed and released.
Or just make the changes yourself, I think both of you are able to push to this branch.

test/dom-parser.test.js
karfau: I don't think it's such big of a difference, but hey, I applied your suggestion.

kboshold (approved):

Ponynjaa (approved):

PR #858, opened 2025-03-14
Sync

logerzhu:


karfau:
We will of course not land a change that changes the name of the library to your scope.
Regarding the other changes it would be great to first know exactly what problem you are solving with it and have a reproduction for it, so that we can add a test for it.

logerzhu:
> We will of course not land a change that changes the name of the library to your scope. Regarding the other changes it would be great to first know exactly what problem you are solving with it and have a reproduction for it, so that we can add a test for it.

Could you help to close this pull request? I am trying to pull to latest code to my fock for test but make a mistake to create this request.

-----------------

While using xmldom to handle large XML of  DOCX file (tens of MB), I encountered an "Invalid array length" error due to the buf array exceeding the maximum supported length. Therefore, I modified the code in dom.js to avoid this error.

source code:
`			while (child) {
				serializeToString(child, buf, nodeFilter, visibleNamespaces.slice());
				child = child.nextSibling;
			}
`
update to
`
                         while (child) {
				//Avoid Invalid array length
				var newBuf = []
				serializeToString(child, newBuf, nodeFilter, visibleNamespaces.slice());
				buf.push(newBuf.join(''))
				child = child.nextSibling;
			}
`

But this modification might impact performance, so it may be necessary to conduct some performance regression tests.

karfau:
Thx for the information.
Can you tell me how to get access to a sample of such a huge XML?
I would like to understand the impact.

One option might be to implement the workaround only before adding something to the buffer that would exceed the max length.

Another option would be to add a constructor option to opt in to the potentially slower implementation.

I would recommend creating an issue referencing this PR. So we can discuss the options and implementation before going for it.

Issue #932, opened 2025-08-17
ownerDocument not updated when appendChild moves nodes between documents


shunkica:
**Describe the bug**
The `ownerDocument` property is not updated when `appendChild()` moves an element from one document to another.

**To Reproduce**
https://stackblitz.com/edit/js-xmldom-template-ma7jzuzj?file=index.js

**Expected behavior**
The `ownerDocument` must be updated to the new document (including the `ownerDocument` of all the children)

**Runtime & Version:**
xmldom version: 0.8.x 0.9.x
runtime version: node v20
other related software and version: 

**Additional context**

PR #933, opened 2025-08-17
fix: [0.8.x] update `ownerDocument` when moving nodes between documents

shunkica:
Resolves issue #932


karfau:
Thank you for your continued contributions, especially the great test coverage. 

Do I understand correctly that this is an issue in 0.8.x and 0.9.x?
If so, I usually first land the fix on master and afterwards apply it to the release-0.8.x branch.
Are you also planning to provide a PR against the master branch?

But since you did it the other way around I took the opportunity to turn on GitHub checks for 0.8.x versions.

karfau (approved):

PR #1038, opened 2026-05-29
reject charrefs to invalid XML characters in entityReplacer

jmestwa-coder:
entityReplacer decodes numeric character references by passing the parsed value straight to fixedFromCharCode without checking it against the XML Char production, so `&#0;`, `&#xB;`, lone surrogates like `&#xD800;` and out-of-range values like `&#x110000;` end up as control characters or stray surrogate code units in the DOM. Validate the code point inside entityReplacer and report an error leaving the reference unexpanded, the way an unknown named entity is already handled. The updated error snapshots only reflect the new reporting call site.

### DIFF
--- a/lib/sax.js
+++ b/lib/sax.js
@@ -66,6 +66,18 @@ function parse(source, defaultNSMapCopy, entityMap, domBuilder, errorHandler) {
 		}
 	}
 
+	function isValidXmlCodePoint(code) {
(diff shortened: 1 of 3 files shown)


jmestwa-coder:
pushed a fix for the failing test job. the not-wf snapshots (142-146) needed updating to reflect the now-rejected invalid charrefs.

PR #1039, opened 2026-05-30
fix: avoid quadratic backtracking in PI regex

jmestwa-coder:
A processing instruction with no closing `?>` makes the `PI` grammar regex backtrack quadratically, since the leading `S+` and the trailing `Char*?` both match whitespace. Reachable from `DOMParser.parseFromString` with default options (`<a><?p` followed by a run of spaces):

    8KB:   359ms -> 2ms
    16KB: 2043ms -> 1ms
    32KB: 7777ms -> 1ms

The non-capturing `(?!S)` after `S+` forces the data to begin at the first non-space character, so the separator can no longer redistribute whitespace into the lazy `Char*?`. Data is still capture group 2.

Fixes the ReDoS reported in [**GHSA-g53g-w8rj-fmg7**](https://github.com/xmldom/xmldom/security/advisories/GHSA-g53g-w8rj-fmg7) (0.9.x only — the 0.8.x line and the unscoped `xmldom` package parse PIs via a different, `indexOf('?>')`-bounded path and are not affected).


jmestwa-coder:
gentle ping

karfau (approved):
Thx for your contribution

PR #1043, opened 2026-06-02
fix: reject invalid element and attribute names when requireWellFormed

jmestwa-coder:
The requireWellFormed serializer validates Comment, ProcessingInstruction, CDATASection, Text and DocumentType content against their XML productions, but never the element or attribute name. createElement and setAttribute do no Name validation, so createElement('x><inject') or setAttribute('a><inject ', v) serialize to injected markup under { requireWellFormed: true } instead of throwing. Reject names that are not a valid QName, the same check createAttribute already applies at creation time.

The check is opt-in (only under `{ requireWellFormed: true }`); default serialization is unchanged. It also covers the **namespace-prefix** sub-vector: an invalid prefix surfaces in the element qualified name (`PREFIX:local`) or in a synthesized `xmlns:PREFIX` declaration, both of which are QName-checked — a regression test exercises both paths. The `serializeToString` `@throws` documentation is synchronized between `lib/dom.js` and `index.d.ts`.

Fixes:
- [**GHSA-w2rr-34g9-rvrj**](https://github.com/xmldom/xmldom/security/advisories/GHSA-w2rr-34g9-rvrj) — element name injection via `createElement()`
- [**GHSA-4w3w-2rp5-g8jm**](https://github.com/xmldom/xmldom/security/advisories/GHSA-4w3w-2rp5-g8jm) — attribute name injection via `setAttribute()`

Creation-time name validation (rejecting invalid names in `createElement`/`setAttribute`) is a breaking change deferred to the next breaking release.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1586,13 +1586,19 @@ declare module '@xmldom/xmldom' {
 		 * concatenated CDATA sections — **deprecated**, will be removed in the next breaking
 		 * release.
 		 * - W3C DOM Parsing §3.2.1.1 requires well-formedness checks on Element `localName`s,
-		 * prefixes, and attribute serialization when `requireWellFormed` is `true`. These checks are
-		 * **not implemented** in this release — see the tracking issue filed against the next
-		 * breaking milestone.
+		 * prefixes, and attribute serialization when `requireWellFormed` is `true`. Element and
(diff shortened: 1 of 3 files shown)


jmestwa-coder:
any update?

karfau (approved):
Thx for your contribution

PR #1044, opened 2026-06-03
check attribute value against char production when requireWellFormed

jmestwa-coder:
Noticed the requireWellFormed serializer checks text, comment, PI, cdata and doctype content against the Char production but never attribute values. addSerializedAttribute emits raw NUL/control chars and returns without throwing, while the same data in a text node already throws. Added the InvalidChar check there so it also covers namespace declarations and directly serialized Attr nodes.


jmestwa-coder:
any update?

